PR #4, opened 2021-05-13
Get blob value bytes by value id

### LINKED ISSUE #3: Get blob value bytes by value id
### Description

Blob values coming back from the server in json aren't populated with the byte array of the file.  Need a function to fetch blob values from the database.

brianbolt:
Fixes #3


bffrost (approved):
LGTM

PR #6, opened 2021-05-28
Allow passing of label_type and label_kind to get_thing_codes_by_labels and get_saved_entity_codes

### LINKED ISSUE #5: Allow passing of label_type and label_kind to get_thing_codes_by_labels and get_saved_entity_codes
The underlying ACAS Base route allows for the optional parameters labelType and labelKind which can really help scope down the query. This helps both with performance and with cases where there are multiple matches for a label but if we filtered by labelType + labelKind we would not see multiple matches.

However the current ACASclient function  doesn't enable passing those parameters.

brianbolt:
Fixes #5

### DIFF
--- a/acasclient/acasclient.py
+++ b/acasclient/acasclient.py
@@ -920,14 +920,16 @@ class client():
         resp.raise_for_status()
         return resp.json()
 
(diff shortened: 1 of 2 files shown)


bffrost (approved):
LGTM

PR #9, opened 2021-06-09
Simple LsThings

bffrost:
- Copied over the whole LsThing / SimpleLsThing / SimpleLink python data classes
- Made breaking change of DDictValue -> CodeValue since that more accurately represents what the class does (as it was not a true DDictValue) and to prevent confusion.
- Removed dead / unused code
- Added a lot of docstrings with examples

I have not yet added any tests. I will try to add some basic tests but there is a substantial amount covered by this new code so we will not have anywhere close to 100% test coverage.

### DIFF
--- a/acasclient/acasclient.py
+++ b/acasclient/acasclient.py
@@ -233,7 +233,8 @@ class client():
                 filesToUpload[str("file")] = base64.decodebytes(file.encode())
             elif isinstance(file, dict):
                 if isBase64(file['data']):
-                    filesToUpload[file["name"]] = base64.decodebytes(file["data"].encode())
+                    filesToUpload[file["name"]] = base64.decodebytes(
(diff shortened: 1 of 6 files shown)


bffrost:
@brianbolt  I'm open to suggestions or opinions you have about splitting this up into multiple files or integrating it better into the package structure

brianbolt:
Thanks for putting in this work.  I'll take a spin through and work on some things and pull request into this branch.

Things off the top of my head are:

- Add requirements so the package can auto build/install
- Move the code over to a "models" directory

I'll play around with it and see what else comes up.

brianbolt:
I ran auto flake 8, fixed some old tests, adding requirements and added a simple project creation and save test.  We can modify it more later but this LGTM.

PR #10, opened 2021-06-15
Things, adding the ability to save blob values and fetch blob value c…

brianbolt:
…ontents by id

### DIFF
--- a/acasclient/lsthing.py
+++ b/acasclient/lsthing.py
@@ -591,6 +591,10 @@ class BlobValue(object):
         :type value: bytes, optional
         :param comments: Filename as a string, defaults to None
         :type comments: str, optional
+        :param id: id as an int, defaults to None
+        :type id: int, optional
+        :param ls_value: ls_value as a <acasclient.lsthing.LsThingValue>, defaults to None
+        :type ls_value: <acasclient.lsthing.LsThingValue>, optional
         """
         if ls_value is not None:
             value = ls_value.blob_value
@@ -613,6 +617,11 @@ class BlobValue(object):
         self.id = id
 
     def get_data(self, client):
+        """Get blob value data as bytes
+
+        :return: bytes of blob value from server
+        :rtype: bytes
+        """
         return client.get_blob_data_by_value_id(self.id)
 
     def __eq__(self, other: object) -> bool:
--- a/acasclient/lsthing.py
+++ b/acasclient/lsthing.py
@@ -6,6 +6,7 @@ import hashlib
 import json
 import logging
 import pathlib
+from pathlib import Path
 import re
 from collections import defaultdict
 from datetime import datetime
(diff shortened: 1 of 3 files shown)


acasclient/lsthing.py
```diff
@@ -584,16 +584,45 @@ class BlobValue(object):
...
+            comments = ls_value.comments
+            id = ls_value.id
+        else:
+            if value is not None:
+                if isinstance(value, Path) or isinstance(value, str):
+                    if isinstance(value, str):
+                        value = Path(value)
+                    if comments is None:
+                        comments = value.name
+                    f = value.open('rb')
```
bffrost, line 610: Should test that this is a valid path to fail gracefully.

acasclient/lsthing.py
```diff
@@ -584,16 +584,45 @@ class BlobValue(object):
...
+        :param ls_value: ls_value as a <acasclient.lsthing.LsThingValue>, defaults to None
+        :type ls_value: <acasclient.lsthing.LsThingValue>, optional
         """
+        if ls_value is not None:
+            value = ls_value.blob_value
+            comments = ls_value.comments
+            id = ls_value.id
+        else:
+            if value is not None:
+                if isinstance(value, Path) or isinstance(value, str):
```
bffrost, line 605: It feels "easter-eggy" to overload the "value" field this way. I'd prefer this as a distinct argument, maybe `filepath` that will set `value` to the bytes and set `comments` to the file name

acasclient/lsthing.py
```diff
@@ -584,16 +584,45 @@ class BlobValue(object):
...
+                    f = value.open('rb')
+                    bytes_array = f.read()
+                    value = [x for x in bytes_array]
+                    f.close()
+
         self.value = value
         self.comments = comments
+        self.id = id
+
+    def get_data(self, client):
```
bffrost, line 619: nit: `get_data` is a little vague and doesn't clearly say that this is fetching from the ACAS server versus just fetching a value from local memory.
Perhaps `get_blob_data_from_acas` if we want to be verbose, or `download_data` ?

acasclient/lsthing.py
bffrost, line 610: Can we also add a helper method to write the byte `value` out to a file?
`def write_to_file(filePath=None):` where it will default to writing to the local directory and naming the file based on `comments` but can be overridden to write out to a specific path?

bffrost (changes requested):

acasclient/lsthing.py
brianbolt: I like `download_data`.  I'll go with that.

bffrost (approved):
Sorry for the delay on the review, but this looks great! Thanks for all the tests and the nice file writing features.

brianbolt:
This is released in v0.1.20 of the acasclient

PR #11, opened 2021-06-16
Add jsonifying of  BlobValue.

anand-schrodinger:
Add logic for jsonifying BlobValue.

Note: I was not not able to run all tests probably because of my setup so I just ran `TestBlobValue` and it passed.


bffrost (approved):
LGTM

PR #12, opened 2021-06-24
Update get_by_code to pass LsThing.

anand-schrodinger:
- Add `is compatible with` interaction ls_type.
- Updated `get_by_code` to create LsThing from camel cased dict keys and pass that in `ls_thing`
I am not sure about the output dict key case of [get_ls_thing](https://github.com/mcneilco/acasclient/blob/183596bb26a68bd36a71234acaae988a1766e10d/acasclient/acasclient.py#L776) but I checked that when we save it from [save](https://github.com/mcneilco/acasclient/blob/ef9e14300c1a2d1cf109b93aeea73a9c1ff713fb/acasclient/lsthing.py#L1180), we change it to camel case.

### DIFF
--- a/acasclient/interactions.py
+++ b/acasclient/interactions.py
@@ -27,7 +27,8 @@ INTERACTION_VERBS_DICT = {
     'comprises': 'is comprised of',
     'analyzes': 'is analyzed by',
     'is input structure for': 'has input structure',
-    'owns': 'is owned by'
+    'owns': 'is owned by',
+    'is compatible with': 'is compatible with',
 }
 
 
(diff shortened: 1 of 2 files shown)


devendrachitturi:
LGTM

devendrachitturi (approved):

acasclient/interactions.py
```diff
@@ -27,7 +27,8 @@
     'comprises': 'is comprised of',
     'analyzes': 'is analyzed by',
     'is input structure for': 'has input structure',
-    'owns': 'is owned by'
+    'owns': 'is owned by',
+    'is compatible with': 'is compatible with',
```
bffrost, line 31: I know I came up with this, but I hate this particular interaction because it is reversible...
TODO for me: think of a better, more specific, directional term and replace this one.

bffrost (approved):

PR #13, opened 2021-07-02
Fix 'make_ls_value' API

anand-schrodinger:
`make_ls_value` should have `BlobValue` as an `elif` statement, else `FileValue` attribute would be overwritten with `str` logic since `FileValue` is a subclass of `str`.


bffrost (approved):
LGTM

devendrachitturi (approved):
LGTM

PR #14, opened 2021-07-07
Add 'client' parameter to SimpleLsThing.

anand-schrodinger:
Added a new `client` argument which stores reference to the acasclient. Intended purpose is to use the client for downloading/uploading information from ACAS.

### DIFF
--- a/acasclient/lsthing.py
+++ b/acasclient/lsthing.py
@@ -682,6 +682,8 @@ class BlobValue(object):
         return full_file_path
 
     def __eq__(self, other: object) -> bool:
+        if other is None:
+            return False
         return self.value == other.value and self.comments == other.comments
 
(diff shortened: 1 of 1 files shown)


bffrost (commented):
I think we should extend this PR. Instance methods like `save`, `_prepare_for_save` and `upload_file_values` could all use `self._client` instead of requiring a `client` argument.

Class methods like `save_list` and `update_list` could also call `set_client` on the returned objects.

bffrost:
If `_prepare_for_save` is going to have a `self._client` we should also have the `user` argument default to `self._client.username` so we seamlessly use the username of whoever is authenticated as the "recorded_by"

bffrost (approved):
I think this PR is good as-is. I'll spin off my suggestions into separate PRs.

PR #17, opened 2021-07-19
Fixes #16 blob value comparison

### LINKED ISSUE #16: BlobValue comparison in `is_equal_ls_value_simple_value` is broken
* acasclient version: v0.1.20
* Python version: 3.8
* Operating System: Ubuntu Linux

### Description

Trying to update an LsThing with a new BlobValue using SimpleLsThing / `_prepare_for_save`. The comparison between old and new BlobValues is broken.

### What I Did

Call `_prepare_for_save` for a SimpleLsThing containing a BlobValue, when `self._ls_thing` also contained an older BlobValue of the same lsKind.
```
 Traceback (most recent call last):
...
   File "/usr/local/lib/python3.8/dist-packages/acasclient/lsthing.py", line 1781, in _prepare_for_save
     metadata_ls_states = update_ls_states_from_dict(
   File "/usr/local/lib/python3.8/dist-packages/acasclient/lsthing.py", line 377, in update_ls_states_from_dict
     ls_values = update_ls_values_from_dict(
   File "/usr/local/lib/python3.8/dist-packages/acasclient/lsthing.py", line 452, in update_ls_values_from_dict
     if not is_equal_ls_value_simple_value(old_ls_val, val_value):
   File "/usr/local/lib/python3.8/dist-packages/acasclient/lsthing.py", line 214, in is_equal_ls_value_simple_value
     return ls_value.blob_value == BlobValue(val.value, val.comments)
   File "/usr/local/lib/python3.8/dist-packages/acasclient/lsthing.py", line 685, in __eq__
     return self.value == other.value and self.comments == other.comments
 AttributeError: 'list' object has no attribute 'value'
```

bffrost:
BlobValue comparison in `is_equal_ls_value_simple_value` was backwards in that it was taking attributes from a BlobValue and re-casting them again as a BlobValue, rather than taking attributes from an LsValue and forming a BlobValue.

Added a test of updating a blobValue to confirm the fix.

### DIFF
--- a/acasclient/lsthing.py
+++ b/acasclient/lsthing.py
@@ -211,7 +211,7 @@ def is_equal_ls_value_simple_value(ls_value, val):
     elif isinstance(val, FileValue):
         return ls_value.file_value == val
     elif isinstance(val, BlobValue):
-        return ls_value.blob_value == BlobValue(val.value, val.comments)
+        return  BlobValue(ls_value.blob_value, ls_value.comments) == val
     elif isinstance(val, clob):
         return ls_value.clob_value == str(val)
     elif type(val) == str:
--- a/tests/test_lsthing.py
+++ b/tests/test_lsthing.py
@@ -166,6 +166,52 @@ class TestLsThing(unittest.TestCase):
             newProject = Project(recorded_by=self.client.username, **meta_dict)
         except ValueError as err:
             self.assertIn("not a file", err.args[0])
+    
+    def test_002_update_blob_value(self):
+        """Test saving simple ls thing with blob value, then updating the blobValue."""
+        # Helpers
+        def _get_path(file_name):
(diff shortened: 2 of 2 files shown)


brianbolt:
Small nit is that the helper functions are built into the test, I tend to think if you are going to make a function you should make it available outside of the test or just make the code inline as it just makes the code a little harder to walk through.  Not a big deal, LGTM.

brianbolt:
Actually, ignore what I just said, I see you are reusing it a few times in the test so its definitely worth being a function. Just wonder if its useful outside this test but not reason to move it outside if it's not currently being used.

bffrost:
@brianbolt Thanks for the feedback. Yeah I made it a function to reuse a few times in the test. But you are right that it could be useful outside the test, so I've moved it and refactored the usages in the other test too.

PR #18, opened 2021-07-19
SimpleLsThing `client` extension

bffrost:
Followup to #14 , implementing my opinionated suggestions. @brianbolt @anand-schrodinger I would like your honest feedback about whether you agree with this direction or not, since this does include breaking changes to some interfaces.

- Make `client` a *required* parameter for instantiating SimpleLsThing
- Remove `client` from all other instance methods since it can now be guaranteed to exist as `self.client`
- Switch `recorded_by` and `modified_by` to default to the active client session's username, unless overridden.
- Added and fixed some tests


brianbolt:
I like the concept but would like to see this change without requiring an active connection to an acas server.  I can imagine wanting to use, test and work with SimpleLSThings without an active ACAS connection.  `acasclient` currently requires being online when being instantiated.  I think if we change acasclient to:

A) not require login when being instantiated (maybe just log a warning that its offline?)  also log errors when trying to call save when offline, then this would work more naturally.

anand-schrodinger (commented):
I like these changes a lot for for how I use these APIs. Thank you!

bffrost:
@brianbolt I've pushed up changes to implement a crude offline mode along the lines of what you were asking. I think to do something more functional we'd need to mock responses for lots of routes, so for now I chose to just raise an exception if anyone tries to call a method that reaches out to ACAS.

What do you think?

PR #19, opened 2021-07-29
ACAS-124: Update 'advanced_search_ls_things' API.

anand-schrodinger:
Add `combine_terms_with_and` and `format` arguments to `advanced_search_ls_things` API.

### DIFF
--- a/tests/test_acasclient.py
+++ b/tests/test_acasclient.py
@@ -1118,6 +1118,17 @@ class TestAcasclient(unittest.TestCase):
                                        max_results=1000)
         self.assertIn(codes[0], return_codes)
 
+        # Use 'codetable' data 'format'; 'codes_only' is False
+        results = self.client\
+            .advanced_search_ls_things('project', 'project', 'active',
+                                       value_listings=value_listings,
+                                       codes_only=False,
+                                       max_results=1000,
+                                       combine_terms_with_and=True,
+                                       format='codetable')
+        return_codes = [res['code'] for res in results]
+        self.assertIn(codes[0], return_codes)
+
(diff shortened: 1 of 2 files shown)


tests/test_acasclient.py
```diff
@@ -1118,6 +1118,17 @@ def test_032_advanced_search_ls_things(self):
                                        max_results=1000)
         self.assertIn(codes[0], return_codes)
 
+        # Use 'codetable' data 'format'; 'codes_only' is False
+        results = self.client\
+            .advanced_search_ls_things('project', 'project', 'active',
+                                       value_listings=value_listings,
+                                       codes_only=False,
+                                       max_results=1000,
+                                       combine_terms_with_and=True,
```
bffrost, line 1127: This test is fine. But a more representative test would be to replace `'active'` with `None` in this call, and instead move `active` into the value listings like so:
```
value_listings[0]["value"] = 'active'
```

It'll be doing the same search but if you expanded out to multiple terms to query:
project status : active
is restricted: false

Then you would want the `'active'` and `'false'` terms to be tied specifically to their respective values.

bffrost (approved):
Approving since the code looks right, but would prefer you add the additional test I recommended.

tests/test_acasclient.py
anand-schrodinger: Thanks! I have updated tests to cover `combine_terms_with_and` too.

anand-schrodinger:
@bffrost @brianbolt I'll need a version bump for these changes to be available. Let me know if there are documented steps that I can follow to do it myself. Thanks!

PR #20, opened 2021-07-29
Switch FileValues to handle comments

bffrost:
Overly simplistic FileValue-as-a-single-string was losing information when an uploaded filename doesn't match the original file name. This happens when reuploading the same file repeatedly, where each file gets ` (1)`, ` (2)`, etc. added into the filename.

This change goes in tandem with https://github.com/mcneilco/acas/pull/790#event-5081839321 to not lose this same information in ACAS's `ACASFormLSFileValueFieldController`


bffrost:
I hit some failures caused by running my tests locally against an `acas_custom` branch that didn't have project defined in `ControllerRedirectConf.coffee`. Adding a configuration for project fixed the issue.
I noted that the default base ACAS ControllerRedirectConf has project already defined: https://github.com/mcneilco/acas/blob/release/1.13.6/modules/ServerAPI/src/server/ControllerRedirectConf.coffee

acasclient/lsthing.py
```diff
@@ -570,15 +570,66 @@ class clob(str):
...
+        acas_file = client.get_file(f'/dataFiles/{self.value}')
+        if self.comments:
+            acas_file["name"] = self.comments
+        return str(client.write_file(acas_file, folder_path))
+    
+    def as_dict(self) -> Dict[str, Any]:
+        """
+        Return a map of attribute name and attribute values stored on the
+        instance.
+        Note: Only attributes stored in `BlobValue._fields` will be returned.
```
anand-schrodinger, line 627: `BlobValue` > `FileValue`.

anand-schrodinger (approved):

PR #22, opened 2021-08-04
ACAS-127: upload_file_values as part of save

bffrost:
- Add `upload_file_values` to `SimpleLsThing` `save`, `save_list`, and `update_list`. Developers should be able to forget this and have acasclient handle this for them transparently.
- Fixed bug in recent FileValue addition. Should have been referencing `file_path` but was referencing `value` instead. Added improved tests to catch this case.


anand-schrodinger (approved):
LGTM!
I like where this is headed. Its now a much cleaner API I guess.

PR #24, opened 2021-08-17
ACAS-132: Expose labels and interactions in advanced_search_ls_things

bffrost:
Only added a test for Labels because it would take much more effort to create interactions to query.


bffrost:
Added tests for interactions too

PR #25, opened 2021-09-01
SS-35238: upload_file_values on saved entities fail

bffrost:
The ill-fated PR #22 introduced several bugs. One is that trying to update an existing LsThing which contains a FileValue (even if one does not intend to edit that filevalue) will fail trying to re-upload the file value.

This fix is not the most elegant - just skipping the attempted upload if the targeted file doesn't exist locally. It could cause intentional uploads to fail silently rather than loudly if the local filepath is wrong.

Open to suggestions or ideas of better ways to do this.


anand-schrodinger (approved):
Do you think if the file doesn't exist we should print out a message -
`Skipping foo.png, file not found`?

bffrost:
I thought about that, but didn't want to spam the logs every time we do an update.
Could make this an option: `upload_file_values(client, ignore_missing_files=False)` and then change the `update` methods to set `ignore_missing_files=True` so initial save will still require files but update won't.

bffrost:
@anand-schrodinger @brianbolt  I've made some major changes to this PR. Could you give another review?
- Moved file uploading logic down to `update_ls_values_from_dict` within `_prepare_for_save` so it only applies to new FileValues. Stopped calling `upload_file_values` during save.
- Added several new unit tests and got them passing

brianbolt (commented):
LGTM

brianbolt (approved):

anand-schrodinger (approved):
Great idea! Tests also look great to me.

Do you think now ` _upload_file_values_from_state_dict` should also just call `_upload_file_value` so that we have a single point of action?

PR #26, opened 2021-09-08
SS-35219: Fixed default argument bug

du-singh:
To understand the issue and fix look at below code.
```
def foo(x, a=[]):
    a.append(x)
    print(a)
    
foo(1)
foo(2)
foo(3)
```
Above code prints below output.
```
[1]
[1, 2]
[1, 2, 3]
```
In python default argument is assigned only once like static variables. Hence, using non-primitive datatypes like list, dict, defaultdict (or some instance foo of class bar) as default arguments can lead to unexpected/undesired behavior.
Fix is simple, see below.
```
def foo(x, a=None):
    a = a or []
    a.append(x)
    print(a)
```

This is causing the duplication of data if default value is of list type or overriding of data when default value is of type dict or defaultdict.
With these changes it should get fixed.

I couldn't test it locally. Will try it by:
1. merging this branch with master
2. bumping acasclient version
3. incrementing acasclient version in server requirements.txt
4. revert these changes if it fails

I wanted to test it locally by using local acasclient code with BE container, but couldn't make it happen. Used this as reference for it https://stackoverflow.com/questions/44708481/how-to-install-local-packages-using-pip-as-part-of-a-docker-build/44708686
Any hints on how to do it?

### DIFF
--- a/acasclient/lsthing.py
+++ b/acasclient/lsthing.py
@@ -1677,7 +1677,6 @@ class SimpleLsThing(BaseModel):
 
     def __init__(self, ls_type=None, ls_kind=None, code_name=None, names=None, ids=None, aliases=None, metadata=None, results=None, links=None, recorded_by=None,
                  preferred_label_kind=None, state_tables=None, ls_thing=None, client=None):
-        metadata = metadata or {}
         self._client = client
         self.preferred_label_kind = preferred_label_kind
         # if ls_thing passed in, just parse from it and ignore the rest
@@ -1694,14 +1693,14 @@ class SimpleLsThing(BaseModel):
             self._ls_thing = LsThing(ls_type=self.ls_type, ls_kind=self.ls_kind,
                                      code_name=self.code_name, recorded_by=self.recorded_by)
             self.names = names or {}
-            self.ids = ids or None
+            self.ids = ids or {}
             self.aliases = aliases or {}
             # Create empty dicts for LsLabels, LsStates, and LsValues
             # These will be populated by the "_prepare_for_save" method
             self._name_labels = {}
             self._id_labels = {}
             self._alias_labels = defaultdict(list)
-            self.metadata = metadata
+            self.metadata = metadata or {}
             self.results = results or {}
             self._metadata_states = {}
             self._metadata_values = {}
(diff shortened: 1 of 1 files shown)


acasclient/lsthing.py
```diff
@@ -1675,8 +1675,9 @@ class SimpleLsThing(BaseModel):
 
     ROW_NUM_KEY = 'row number'
 
-    def __init__(self, ls_type=None, ls_kind=None, code_name=None, names={}, ids={}, aliases={}, metadata={}, results={}, links=[], recorded_by=None,
-                 preferred_label_kind=None, state_tables=defaultdict(dict), ls_thing=None, client=None):
+    def __init__(self, ls_type=None, ls_kind=None, code_name=None, names=None, ids=None, aliases=None, metadata=None, results=None, links=None, recorded_by=None,
+                 preferred_label_kind=None, state_tables=None, ls_thing=None, client=None):
+        metadata = metadata or {}
```
bffrost, line 1680: Confused why this is here and not on line 1704. If there is a reason it seems like `results` and `state_tables` should receive the same treatment.

acasclient/lsthing.py
```diff
@@ -2100,12 +2101,14 @@ class SimpleLink(BaseModel):
...
     _fields = ['verb', 'subject', 'object', 'metadata', 'results']
 
-    def __init__(self, verb=None, subject=None, object=None, metadata={}, results={}, recorded_by=None, itx_ls_thing_ls_thing=None,
+    def __init__(self, verb=None, subject=None, object=None, metadata=None, results=None, recorded_by=None, itx_ls_thing_ls_thing=None,
                  subject_type=None, object_type=None):
         """
         Create a link of form: "{subject} {verb} {object}" where {subject} and {object} are instances of SimpleLsThing
         examples: "{batch} {instantiates} {parent}", "{literature reference} {contains} {pdb structure}"
         """
+        metadata = metadata or {}
```
bffrost, line 2110: Similar question here - is there a reason this is at the top and not on line 2153?

bffrost (commented):
Left some questions. Have you run the ldclient unit tests? They require a running local instance of ACAS. Running at least the `LsThing` tests would make me feel more confident there isn't something obvious we've missed.

After that, we can proceed with the test plan as you wrote it.

acasclient/lsthing.py
du-singh: I did that to follow DRY
```
metadata = metadata or {}
foo = metadata
bar = fn(metadata)
```
vs 
```
foo = metadata or {}
bar = fn(metadata or {})
```
Two functions use metadata more than once, but others use it once only (hence direct modification for those).
I saw your commit, now below places will pick `None` as `metadata` value instead of `{}`
https://github.com/mcneilco/acasclient/blob/master/acasclient/lsthing.py#L2153
https://github.com/mcneilco/acasclient/blob/master/acasclient/lsthing.py#L1691

acasclient/lsthing.py
du-singh: Same as above

acasclient/lsthing.py
bffrost: I see. My mistake - sorry, and thanks for the fix.

bffrost: Sorry for the bad commit. Thanks for the fix.

PR #27, opened 2021-09-13
SS-35219: Fixed default value for 'metadata' or 'results'

du-singh:
Fixed the issue I mentioned [here](https://github.com/mcneilco/acasclient/pull/26#discussion_r706571210).


bffrost:
Sorry for the bad commit. Thanks for fixing here.

PR #30, opened 2021-10-12
Ability to specify the subject_type and object_type when adding links

### LINKED ISSUE #29: Ability to specify the subject_type and object_type when adding links
* acasclient version: 0.1.28
* Python version: 3.8
* Operating System: NA

### Description

Trying to call SimpleLsThing.add_link but specify the ls_kind of the interaction however add_link just combines the {first_thing.ls_type}_{second_thing.ls_type} to create the interaction kind.  ItxLsThingLsThing should have the ability to accept alternative subject_type and object_type strings and SImpleLsThing.add_link should be able to pass those arguments to ItxLsThingLsThing.

brianbolt:
Fixes #29

### DIFF
--- a/acasclient/lsthing.py
+++ b/acasclient/lsthing.py
@@ -2028,7 +2028,7 @@ class SimpleLsThing(BaseModel):
                 buf = file_ref.read(BLOCKSIZE)
         return(hasher.hexdigest())
 
-    def add_link(self, verb=None, linked_thing=None, recorded_by=None, metadata=None, results=None, **kwargs):
+    def add_link(self, verb=None, linked_thing=None, recorded_by=None, metadata=None, results=None, subject_type=None, **kwargs):
         """Create a new link between this SimpleLsThing and another SimpleLsThing `linked_thing`
 
         :param verb: The nature of the link. This should be defined in `interactions.py`, defaults to None
@@ -2042,8 +2042,10 @@ class SimpleLsThing(BaseModel):
         :param results: Dictionary of results to associate with the link itself, defaults to {}
         :type results: dict, optional
         """
+        if not subject_type:
+            subject_type = self.ls_type
         self.links.append(SimpleLink(verb=verb, object=linked_thing, recorded_by=recorded_by, metadata=metadata or {},
-            results=results or {}, **kwargs))
+            subject_type=subject_type, results=results or {}, **kwargs))
 
     def upload_file_values(self, client):
         """Loop through the values for file values and check if the value is a base64 string or
--- a/acasclient/lsthing.py
+++ b/acasclient/lsthing.py
@@ -2041,6 +2041,8 @@ class SimpleLsThing(BaseModel):
         :type metadata: dict, optional
         :param results: Dictionary of results to associate with the link itself, defaults to {}
         :type results: dict, optional
+        :param subject_type: The type of the subject of the link, defaults to the ls_type of itself
+        :type subject_type: str, optional
         """
         if not subject_type:
             subject_type = self.ls_type
--- a/acasclient/lsthing.py
+++ b/acasclient/lsthing.py
@@ -2028,7 +2028,7 @@ class SimpleLsThing(BaseModel):
                 buf = file_ref.read(BLOCKSIZE)
         return(hasher.hexdigest())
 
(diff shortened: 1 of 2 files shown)


acasclient/lsthing.py
```diff
@@ -2028,7 +2028,7 @@ def get_file_hash(self, file_path):
                 buf = file_ref.read(BLOCKSIZE)
         return(hasher.hexdigest())
 
-    def add_link(self, verb=None, linked_thing=None, recorded_by=None, metadata=None, results=None):
+    def add_link(self, verb=None, linked_thing=None, recorded_by=None, metadata=None, results=None, subject_type=None, **kwargs):
```
brianbolt, line 2031: add subject_type as param, and **kwargs to pass to SimpleLink. This allows us to override subject_type and object_type in SImpleLink

acasclient/lsthing.py
```diff
@@ -2039,11 +2039,15 @@ def add_link(self, verb=None, linked_thing=None, recorded_by=None, metadata=None
...
-        :param results: Dictioanry of results to associate with the link itself, defaults to {}
+        :param results: Dictionary of results to associate with the link itself, defaults to {}
         :type results: dict, optional
+        :param subject_type: The type of the subject of the link, defaults to the ls_type of itself
+        :type subject_type: str, optional
         """
-        self.links.append(SimpleLink(verb=verb, object=linked_thing, subject_type=self.ls_type,
-            recorded_by=recorded_by, metadata=metadata or {}, results=results or {}))
+        if not subject_type:
+            subject_type = self.ls_type
```
brianbolt, line 2048: subject_type continues to default to the current objects ls_type if not specified as a param

acasclient/lsthing.py
```diff
@@ -2170,12 +2174,14 @@ def __init__(self, verb=None, subject=None, object=None, metadata=None, results=
                 if subject:
                     first_ls_thing = subject._ls_thing
                     first_type = subject.ls_type
-                else:
+                # If subject_type is provided, use it instead of the subject's ls_type
+                if subject_type:
```
brianbolt, line 2178: We use subject_type if it is passed in as a param instead of only if the subject is not passed in.  This gives more explicit control to the user of SimpleLink to specify the subject_type.  This applies to the object_type as well below.

tests/test_lsthing.py
```diff
@@ -399,6 +399,7 @@ def test_005_create_interactions(self):
         fresh_proj_1 = Project.get_by_code(proj_1.code_name, self.client, Project.ls_type, Project.ls_kind)
         assert len(fresh_proj_1.links) == 1
         assert len(fresh_proj_1._ls_thing.second_ls_things) == 1
+        assert fresh_proj_1._ls_thing.second_ls_things[0].ls_kind == f'{Project.ls_type}_{Project.ls_type}'
```
brianbolt, line 402: Add a test to verify that the ls_type and ls_kind of the objects being passed in are still used as the default

tests/test_lsthing.py
```diff
@@ -445,6 +446,28 @@ def test_005_create_interactions(self):
...
+        proj_1.add_link(FWD_ITX, proj_2, recorded_by=self.client.username, subject_type=subject_type, object_type=object_type)
+        assert len(proj_1.links) == 1
+        # save the interaction
+        proj_1.save(self.client)
+        # Fetch project 1 and look at the interaction
+        fresh_proj_1 = Project.get_by_code(proj_1.code_name, self.client, Project.ls_type, Project.ls_kind)
+        assert len(fresh_proj_1.links) == 1
+        assert len(fresh_proj_1._ls_thing.second_ls_things) == 1
+        assert fresh_proj_1._ls_thing.second_ls_things[0].ls_kind == f'{subject_type}_{object_type}'
+
```
brianbolt, line 470: New test to confirm that the new interaction customization is working properly.

acasclient/lsthing.py
bffrost: Is there a reason why `object_type` is buried in kwargs but `subject_type` is exposed directly?

bffrost (approved):
Code changes look good. I'm curious about the use case for this. What's the driver for making the Interaction Kind directly configurable?
And do we want it to be fully configurable? Right now it will still be in the `{foo}_{bar}` format.

PR #41, opened 2022-03-14
Tests fail due to small variation in curve fit values

### LINKED ISSUE #40: Tests fail due to small variation in curve fit values
### Description
Tests fail when running on intel processor.  Small variations in the calculations of Dose response curve fits cause tests to fail as the number are not rounded.

brianbolt:
Fixes #40

### DIFF
--- a/tests/test_acasclient.py
+++ b/tests/test_acasclient.py
@@ -50,6 +50,15 @@ def remove_common_value(value):
     value["analysisGroupId"] = None
     value["stateId"] = None 
 
+    # Round numeric values to 6 digits for comparison on different systems where calculations like EC50 might be different
+    if value['lsType'] == "numericValue":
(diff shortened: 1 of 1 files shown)


tests/test_acasclient.py
```diff
@@ -209,6 +218,7 @@ def setUp(self):
         creds = acasclient.get_default_credentials()
         self.client = acasclient.client(creds)
         self.tempdir = tempfile.mkdtemp()
+        self.maxDiff = None
```
bffrost, line 221: maxDiff seems to be unused.

brianbolt: Just added a comment, its a unittest thing

bffrost (approved):
Thanks for the explanation. I see this is to diff the full JSON output. LGTM

Issue #42, opened 2022-03-17
New feature: Add automatic validation of CodeValues


bffrost:
### Description

CodeValues in ACAS come from controlled vocabularies, most often vocabularies stored in ACAS's Data Dictionary (DDict) tables. A common task in programming with ACAS is to validate that an incoming value is a valid option within a given data dictionary.

We should add a convenient utility for doing this to acasclient to facilitate rapid development.

PR #43, opened 2022-03-17
ACAS-235: Create acasclient utility for validating CodeValues

bffrost:
## Description
I set out to make a utility that would validate that the codes in CodeValues are actually valid, i.e. they show up in the appropriate ACAS DDict. I quickly found that I'd implemented an early version of this, but it relied on an acasclient session being passed in at the time a CodeValue was instantiated, which basically never happens.

I added some unit tests, then started refactoring:
- *Big change*: Added CodeValue validation into a new `SimpleLsThing.validate` function that now happens as part of `save` unless explicitly opted-out.
- Created a new "DDict" generic interface, with implementations for ACAS DDicts and ACAS LsThing references. These are meant as a place to store fetching and checking logic for each `codeOrigin`.
- I then refactored the existing CodeValue validation logic to instead always go through DDict

## Related Issue
#42 

## How Has This Been Tested?
I've run the LsThing tests locally and they pass. I'm seeing some failures with the ACASClient unit tests that I'm not sure are related to my changes, and might be peculiar to my dev environment.

### DIFF
--- a/acasclient/ddict.py
+++ b/acasclient/ddict.py
@@ -8,6 +8,10 @@ logging.basicConfig(level=logging.INFO)
 
 
 class DDict(object):
+    """The DDict class is meant as a generic interface for any implementation of a Data Dictionary that
+    can be referenced by an ACAS CodeValue.
+    Any classes implementing this interface must implement the get_values() method.
+    """
 
     def __init__(self, code_type, code_kind, code_origin):
         self.code_type = code_type
@@ -25,6 +29,7 @@ class DDict(object):
 
 
 class ACASDDict(DDict):
+    """DDict implementation for built-in ACAS DDicts """
 
     CODE_ORIGIN = 'ACAS DDict'
 
@@ -45,6 +50,7 @@ class ACASDDict(DDict):
         return super(ACASDDict, self).check_value(value)
 
 class ACASLsThingDDict(DDict):
+    """DDict implementation for referencing ACAS LsThings"""
 
     CODE_ORIGIN = 'ACAS LsThing'
 
--- a/acasclient/ddict.py
+++ b/acasclient/ddict.py
@@ -10,7 +10,8 @@ logging.basicConfig(level=logging.INFO)
 class DDict(object):
     """The DDict class is meant as a generic interface for any implementation of a Data Dictionary that
     can be referenced by an ACAS CodeValue.
-    Any classes implementing this interface must implement the get_values() method.
+    Any classes implementing this interface must implement the update_valid_values() method.
+    The `update_valid_values` method must also be called before calling `check_value`
     """
 
     def __init__(self, code_type, code_kind, code_origin):
@@ -19,12 +20,14 @@ class DDict(object):
         self.code_origin = code_origin
         self.valid_values = None
 
-    def get_values(self):
+    def update_valid_values(self, client):
         raise NotImplementedError()
 
     @validation_result
     def check_value(self, value):
         """Check if the value is within `self.valid_values` for the DDict."""
+        if not self.valid_values:
+            raise ValueError('Valid values has not been populated. Call `update_valid_values` first.')
         if value not in self.valid_values:
             return False, f"Invalid 'code':'{value}' provided for the given 'code_type':'{self.code_type}' and 'code_kind':'{self.code_kind}'"
 
@@ -37,12 +40,12 @@ class ACASDDict(DDict):
     def __init__(self, code_type, code_kind):
         super(ACASDDict, self).__init__(code_type, code_kind, self.CODE_ORIGIN)
 
-    def get_values(self, client):
+    def update_valid_values(self, client):
         """Get the valid values for the DDict."""
         valid_codetables = client.get_ddict_values_by_type_and_kind(
             self.code_type, self.code_kind)
         self.valid_values = [val_dict['code'] for val_dict in valid_codetables]
-        if self.valid_values == []:
+        if not self.valid_values:
             raise ValueError(f"Invalid 'code_type':'{self.code_type}' or "
                     f"'code_kind':'{self.code_kind}' provided")
 
@@ -54,7 +57,7 @@ class ACASLsThingDDict(DDict):
     def __init__(self, thing_type, thing_kind):
         super(ACASLsThingDDict, self).__init__(thing_type, thing_kind, self.CODE_ORIGIN)
     
-    def get_values(self, client):
+    def update_valid_values(self, client):
         """Get the valid values for the DDict."""
         valid_codetables = client.get_ls_things_by_type_and_kind(self.code_type, self.code_kind, format='codetable')
         self.valid_values = [val_dict['code'] for val_dict in valid_codetables]
--- a/acasclient/ddict.py
+++ b/acasclient/ddict.py
@@ -1,5 +1,4 @@
 from __future__ import unicode_literals
-from typing import Any, Dict
 
 import logging
 
--- a/acasclient/ddict.py
+++ b/acasclient/ddict.py
@@ -1,4 +1,5 @@
 from __future__ import unicode_literals
+from .validation import validation_result
 
 import logging
 
@@ -21,10 +22,11 @@ class DDict(object):
     def get_values(self):
         raise NotImplementedError()
 
+    @validation_result
     def check_value(self, value):
+        """Check if the value is within `self.valid_values` for the DDict."""
         if value not in self.valid_values:
-            raise ValueError(f"Invalid 'code':'{value}' provided for the given "
-                f"'code_type':'{self.code_type}' and 'code_kind':'{self.code_kind}'")
+            return False, f"Invalid 'code':'{value}' provided for the given 'code_type':'{self.code_type}' and 'code_kind':'{self.code_kind}'"
 
 
 class ACASDDict(DDict):
@@ -36,6 +38,7 @@ class ACASDDict(DDict):
         super(ACASDDict, self).__init__(code_type, code_kind, self.CODE_ORIGIN)
 
     def get_values(self, client):
+        """Get the valid values for the DDict."""
         valid_codetables = client.get_ddict_values_by_type_and_kind(
             self.code_type, self.code_kind)
         self.valid_values = [val_dict['code'] for val_dict in valid_codetables]
@@ -52,5 +55,6 @@ class ACASLsThingDDict(DDict):
         super(ACASLsThingDDict, self).__init__(thing_type, thing_kind, self.CODE_ORIGIN)
     
     def get_values(self, client):
+        """Get the valid values for the DDict."""
         valid_codetables = client.get_ls_things_by_type_and_kind(self.code_type, self.code_kind, format='codetable')
         self.valid_values = [val_dict['code'] for val_dict in valid_codetables]
--- a/acasclient/ddict.py
+++ b/acasclient/ddict.py
@@ -13,16 +13,15 @@ class DDict(object):
         self.code_type = code_type
         self.code_kind = code_kind
         self.code_origin = code_origin
-        self.values = None
+        self.valid_values = None
 
     def get_values(self):
         raise NotImplementedError()
 
     def check_value(self, value):
-        if value in self.valid_values:
-            return True
-        else:
-            return False
+        if value not in self.valid_values:
+            raise ValueError(f"Invalid 'code':'{value}' provided for the given "
+                f"'code_type':'{self.code_type}' and 'code_kind':'{self.code_kind}'")
 
 
 class ACASDDict(DDict):
@@ -36,6 +35,14 @@ class ACASDDict(DDict):
         valid_codetables = client.get_ddict_values_by_type_and_kind(
             self.code_type, self.code_kind)
         self.valid_values = [val_dict['code'] for val_dict in valid_codetables]
+        if self.valid_values == []:
+            raise ValueError(f"Invalid 'code_type':'{self.code_type}' or "
+                    f"'code_kind':'{self.code_kind}' provided")
+    
+    def check_value(self, value, client):
+        if not self.valid_values:
+            self.get_values(client)
+        return super(ACASDDict, self).check_value(value)
 
 class ACASLsThingDDict(DDict):
 
@@ -48,3 +55,7 @@ class ACASLsThingDDict(DDict):
         valid_codetables = client.get_ls_things_by_type_and_kind(self.code_type, self.code_kind, format='codetable')
         self.valid_values = [val_dict['code'] for val_dict in valid_codetables]
     
+    def check_value(self, value, client):
+        if not self.valid_values:
+            self.get_values(client)
+        return super(ACASLsThingDDict, self).check_value(value)
--- a/acasclient/ddict.py
+++ b/acasclient/ddict.py
@@ -42,11 +42,6 @@ class ACASDDict(DDict):
         if self.valid_values == []:
             raise ValueError(f"Invalid 'code_type':'{self.code_type}' or "
                     f"'code_kind':'{self.code_kind}' provided")
-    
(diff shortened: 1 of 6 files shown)


brianbolt (changes requested):
Overall I think it's great to add this type of business logic validation to ACAS. Something we are definitely lacking.  I have some nits and some worries about performance.  I also think test_009 could use some additionally comments on what exactly is being tested.

requirements_dev.txt
```diff
@@ -7,4 +7,5 @@ tox==3.14.0
 coverage==4.5.4
 Sphinx==1.8.5
 twine==1.14.0
+pandas==1.4.1
```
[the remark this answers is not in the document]
bffrost: Thanks for the info. I was just trying to get my virtualenv for running unit tests working. I should've used setup.py. I'm going to revert these changes.

tests/test_lsthing.py
```diff
@@ -702,6 +717,80 @@ def test_007_advanced_search_interactions(self):
...
+            START_DATE_KEY: datetime.now(),
+            DESCRIPTION_KEY: desc_1
+        }
+
+        proj_1 = Project(recorded_by=self.client.username, **meta_dict)
+        # set status to a CodeValue constructed with a DDict
+        STATUS_DDICT = ACASDDict(PROJECT, STATUS)
+        proj_1.metadata[PROJECT_METADATA][STATUS_KEY] = CodeValue(ACTIVE, ddict=STATUS_DDICT)
+        proj_1.validate(self.client)
+        assert True
```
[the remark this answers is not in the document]
bffrost: My idea here was just to confirm that we'd passed the "validate" function without throwing an error, but I think this is a silly way of doing it, so I'm going to remove these.

tests/test_lsthing.py
```diff
@@ -702,6 +717,80 @@ def test_007_advanced_search_interactions(self):
...
+        proj_1.metadata[PROJECT_METADATA][STATUS_KEY] = CodeValue(status_1, ddict=STATUS_DDICT)
+        error = None
+        try:
+            proj_1.validate(self.client)
+        except ValueError as e:
+            error = e
+        self.assertEqual(str(error), f"Invalid 'code':'{status_1}' provided for the given 'code_type':'{PROJECT}' and 'code_kind':'{STATUS}'")
+        # Now try adding a CodeValue that references an LsThing
+        proj_1.metadata[PROJECT_METADATA][STATUS_KEY] = CodeValue(ACTIVE, ddict=STATUS_DDICT)
+        proj_1.save(self.client)
```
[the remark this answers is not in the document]
bffrost: Thanks. This line is actually just setup (saving a new LsThing) so that we can reference it below. I've adopted your suggestion above when I first validate `proj_1` then again below when I validate `proj_2`

acasclient/lsthing.py
```diff
@@ -2076,6 +2085,26 @@ def _upload_file_values_from_state_dict(state_dict):
...
+
+        :param client: Authenticated instance of acasclient.client
+        :type client: acasclient.client
+        """
+        for _, values_dict in self.metadata.items():
+            for _, value in values_dict.items():
+                if isinstance(value, CodeValue):
+                    if value.code:
+                        value.validate(client)
+        for _, values_dict in self.results.items():
```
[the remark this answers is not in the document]
bffrost: Thanks, that's cleaner. Incorporated this idea.

acasclient/lsthing.py
```diff
@@ -2076,6 +2085,26 @@ def _upload_file_values_from_state_dict(state_dict):
...
         self.results = _upload_file_values_from_state_dict(self.results)
+    
+    def validate(self, client):
+        """Validate SimpleLsThing dictionary. Currently only checks that all CodeValues that reference ACAS DDicts
+        have known/valid values.
+
+        :param client: Authenticated instance of acasclient.client
+        :type client: acasclient.client
+        """
+        for _, values_dict in self.metadata.items():
```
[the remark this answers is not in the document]
bffrost: Thanks for the comment, I agree. I've shifted responsibility for fetching the contents of the DDicts up to some SimpleLsThing methods so that this can be done in bulk.

Also, `self.metadata` is the SimpleLsThing dictionary, so it won't have ignored or deleted values. The underlying `_ls_thing` object will, but we're not validating that.

acasclient/lsthing.py
```diff
@@ -2076,6 +2085,26 @@ def _upload_file_values_from_state_dict(state_dict):
             return state_dict
         self.metadata = _upload_file_values_from_state_dict(self.metadata)
         self.results = _upload_file_values_from_state_dict(self.results)
+    
+    def validate(self, client):
```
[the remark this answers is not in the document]
bffrost: I ended up removing client and pushing that responsibility onto the DDict.

tests/test_lsthing.py
```diff
@@ -702,6 +717,80 @@ def test_007_advanced_search_interactions(self):
                                        max_results=1000,
                                        combine_terms_with_and=True)
         assert len(results) == 0
+    
+    def test_008_validate_codevalue(self):
```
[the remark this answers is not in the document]
bffrost: Yep, added better comments here and below.

bffrost (commented):
@brianbolt Thank you for the feedback. I've made a number of changes to incorporate your comments. I'm still not feeling great about the approach of raising Exceptions on the first invalid codevalue - I think the more ACAS-canonical way would be to collect a series of error messages and attributes, then raise an error at the SimpleLsThing level in the final validation.

I'm going to work on that change, but pushing up my interim work so you can review.

acasclient/lsthing.py
```diff
@@ -971,24 +960,40 @@ def _validate_params(self, code, code_type, code_kind, code_origin, client):
         :type code_kind: str
         :param code_origin: Origin of DDictValue referenced, typically 'ACAS DDict'
         :type code_origin: str
```
iamanandkr, line 962: Add documentation for  the new`ddict` argument?

acasclient/lsthing.py
```diff
@@ -971,24 +960,40 @@ def _validate_params(self, code, code_type, code_kind, code_origin, client):
...
+
+    def __hash__(self):
+        return hash(f'{self.code}-{self.code_type}-{self.code_kind}-'
+                    f'{self.code_origin}')
+
+    def validate(self):
+        """Validate that this CodeValue conforms to the stored list of possible DDictValues in self.ddict
+
         :return: Error message, or None if valid
         :rtype: str | None
```
iamanandkr, line 991: `ValueError` will be raised here in case of invalid ddict.

I believe we are going to remove the validation logic raising a `ValueError` exception, so this is optional based on what we end up doing.

acasclient/lsthing.py
```diff
@@ -971,24 +960,40 @@ def _validate_params(self, code, code_type, code_kind, code_origin, client):
...
-            return
-        valid_val_dicts = client.get_ddict_values_by_type_and_kind(
-            code_type, code_kind)
-        if valid_val_dicts == []:
-            return (f"Invalid 'code_type':'{code_type}' or "
-                    f"'code_kind':'{code_kind}' provided")
-        if any([code == val_dict['code'] for val_dict in valid_val_dicts]):
+        if self.ddict:
+            self.ddict.check_value(self.code)
             return
```
iamanandkr, line 995: Not required.

acasclient/lsthing.py
```diff
@@ -2076,6 +2109,62 @@ def _upload_file_values_from_state_dict(state_dict):
...
+        for state_dict in state_dicts:
+            for _, values_dict in state_dict.items():
+                for _, value in values_dict.items():
+                    if isinstance(value, CodeValue):
+                        ddict = None
+                        if value.code_origin.upper() == ACAS_DDICT:
+                            ddict = ACASDDict(value.code_type, value.code_kind)
+                        elif value.code_origin.upper() == ACAS_LSTHING:
+                            ddict = ACASLsThingDDict(value.code_type, value.code_kind)
+                        ddicts[(ddict.code_type, ddict.code_kind, ddict.code_origin.upper())] = ddict
```
iamanandkr, line 2128: If the code_origin doesn't match ACAS_DDICT or ACAS_LSTHING, we will here raise an `AttributeError: NoneType object has no attribute ...`. Do you think we should raise a `ValueError` with custom message?

acasclient/lsthing.py
```diff
@@ -2076,6 +2109,62 @@ def _upload_file_values_from_state_dict(state_dict):
...
         self.results = _upload_file_values_from_state_dict(self.results)
+    
+    def _get_ddicts(self):
+        """Extract a unique list of DDict objects from the CodeValue attributes of a SimpleThing
+        :return: dictionary of { (code_type, code_kind, code_origin): DDict }
+        """
+        ddicts = {}
+        state_dicts = [self.metadata, self.results]
+        for state_dict in state_dicts:
+            for _, values_dict in state_dict.items():
```
iamanandkr, line 2120: You could alternatively do:
```python
for values_dict in state_dict.values():
    ...
```

It could be the same with for-loops where we are ignoring the key.

acasclient/ddict.py
```diff
@@ -0,0 +1,56 @@
...
+import logging
+
+logger = logging.getLogger(__name__)
+logging.basicConfig(level=logging.INFO)
+
+
+class DDict(object):
+    """The DDict class is meant as a generic interface for any implementation of a Data Dictionary that
+    can be referenced by an ACAS CodeValue.
+    Any classes implementing this interface must implement the get_values() method.
```
iamanandkr, line 12: What do you think about documenting that the caller should ensure `get_values` is called before calling `check_value`?

If the caller forgets to call `get_values` and directly calls `check_value`, they might get a `TypeError` because of `valid_values` being `None`.

acasclient/lsthing.py
```diff
@@ -19,6 +21,8 @@
 
 
 ROW_NUM_KEY = 'row number'
+ACAS_DDICT = 'ACAS DDICT'
+ACAS_LSTHING = 'ACAS LSTHING'
```
iamanandkr, line 25: I wonder if we should write this as: 🤔 
```python
ACAS_DDICT = ACASDDict.CODE_ORIGIN.upper()
ACAS_LSTHING = ACASLsThingDDict.CODE_ORIGIN.upper()
```

I was thinking on lines of removing duplication in knowledge.

brianbolt:
@bffrost going to let you work on the new changes and incorporating @iamanandkr feedback before the next review I do.

acasclient/ddict.py
```diff
@@ -0,0 +1,67 @@
...
+                f"'code_type':'{self.code_type}' and 'code_kind':'{self.code_kind}'")
+
+
+class ACASDDict(DDict):
+    """DDict implementation for built-in ACAS DDicts """
+
+    CODE_ORIGIN = 'ACAS DDict'
+
+    def __init__(self, code_type, code_kind):
+        super(ACASDDict, self).__init__(code_type, code_kind, self.CODE_ORIGIN)
```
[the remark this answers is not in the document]
bffrost: Yeah - I recall hitting this and realizing it is a python2 vs python3 issue. I think this way I have it is backwards-compatible with python2, so I'm not planning on changing it right now.

acasclient/ddict.py
```diff
@@ -0,0 +1,67 @@
...
+    Any classes implementing this interface must implement the get_values() method.
+    """
+
+    def __init__(self, code_type, code_kind, code_origin):
+        self.code_type = code_type
+        self.code_kind = code_kind
+        self.code_origin = code_origin
+        self.valid_values = None
+
+    def get_values(self):
```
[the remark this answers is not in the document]
bffrost: Agree on having `client` for consistency. I reworked a little how this gets populated so it is getting called from outside this class. In that case I'll rename it to `update_valid_values` since it is a public method.

acasclient/ddict.py
```diff
@@ -0,0 +1,67 @@
...
+    CODE_ORIGIN = 'ACAS DDict'
+
+    def __init__(self, code_type, code_kind):
+        super(ACASDDict, self).__init__(code_type, code_kind, self.CODE_ORIGIN)
+
+    def get_values(self, client):
+        valid_codetables = client.get_ddict_values_by_type_and_kind(
+            self.code_type, self.code_kind)
+        self.valid_values = [val_dict['code'] for val_dict in valid_codetables]
+        if self.valid_values == []:
```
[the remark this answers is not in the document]
bffrost: Thanks, I'll use the pythonic way.

acasclient/ddict.py
bffrost: Great suggestion, adding this both to the docs and adding an explicit exception if someone tries to call `check_value` without `valid_values` populated.

acasclient/lsthing.py
bffrost: Yep, thanks.

bffrost:
Thank you @brianbolt  and @iamanandkr for the reviews. I've addressed your comments and also switched from an Exception-based approach to a `ValidationResult` based approach for handling errors/warnings in validation. If you could both review the new changes I'd really appreciate it!

acasclient/ddict.py
```diff
@@ -0,0 +1,63 @@
...
+        self.code_origin = code_origin
+        self.valid_values = None
+
+    def update_valid_values(self, client):
+        raise NotImplementedError()
+
+    @validation_result
+    def check_value(self, value):
+        """Check if the value is within `self.valid_values` for the DDict."""
+        if not self.valid_values:
```
iamanandkr, line 29: Curious: Do we want to show the custom message if `valid_values` is an empty list or only when `valid_values` is `None`? Checking if we should have this as,
```python
if self.valid_values is None:
    ...
```

acasclient/lsthing.py
```diff
@@ -971,23 +961,43 @@ def _validate_params(self, code, code_type, code_kind, code_origin, client):
...
+    def validate(self):
+        """Validate that this CodeValue conforms to the stored list of possible DDictValues in self.ddict
+
+        :return: ValidationResult indicating True / valid if code is valid, False / invalid otherwise
+        :rtype: ValidationResult
+        """
+        if self.ddict:
+            return self.ddict.check_value(self.code)
+        else:
+            return True
```
iamanandkr, line 1000: Optional: We don't necessarily need it. If the function return value is `None` it will automatically be type cast to `ValidationResult(True, [])`.

iamanandkr (approved):
LGTM! I have added minor optional suggestions/comments, please feel free to ignore them.

acasclient/ddict.py
```diff
@@ -36,18 +40,14 @@ class ACASDDict(DDict):
...
         super(ACASDDict, self).__init__(code_type, code_kind, self.CODE_ORIGIN)
 
-    def get_values(self, client):
+    def update_valid_values(self, client):
+        """Get the valid values for the DDict."""
         valid_codetables = client.get_ddict_values_by_type_and_kind(
             self.code_type, self.code_kind)
         self.valid_values = [val_dict['code'] for val_dict in valid_codetables]
-        if self.valid_values == []:
+        if not self.valid_values:
```
brianbolt, line 48: Just noting that @iamanandkr comment about using `self.valid_values is None` applies here as well

brianbolt (approved):
LGTM

acasclient/ddict.py
bffrost: Right, but here we are actually setting `self.valid_values` on the line above, so I think it will always be either empty list or populated list.

acasclient/ddict.py
bffrost: Yeah, this maybe could be better but in the ACASDDict implementation of `update_valid_values` we throw a specific error if no valid values come back (empty list), so I think this will only be hit when someone calls without first calling `update_valid_values`. Going to merge this as-is.

PR #45, opened 2022-04-12
Tests are saving null "project status"

### LINKED ISSUE #44: Tests are saving null "project status"
* acasclient version: 0.1.32

### Description

Tests save Project ls things with null status which breaks the UI

### What I Did

Ran tests
Tried to save a Lot via the UI and got an error saying there were no projects on the system.  Discovered that this is because Project ls things were being saved with a null `project status` code value.

brianbolt:
## Description
Fixed issue where projects status were being saved with a null value

## Related Issue
Fixes #44

## How Has This Been Tested?
Verified all acasclient tests pass
Cleared the database and verified that this query returns 0 rows where previously several values were added that had null project status:

```
select code_value from ls_thing_value where ls_kind = 'project status' and code_value is null;
```

### DIFF
--- a/tests/test_lsthing.py
+++ b/tests/test_lsthing.py
@@ -350,7 +350,8 @@ class TestLsThing(unittest.TestCase):
 
         # Test updating other values on a saved Thing
         saved_project = Project.get_by_code(newProject.code_name, self.client, Project.ls_type, Project.ls_kind)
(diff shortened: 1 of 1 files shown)


bffrost (approved):
Oops. Thanks for the fix here.

PR #49, opened 2022-04-19
Update acasclient to handle errors & warnings separately

bffrost:
- #43 added the ValidationResult class that handles a boolean "is_valid" status and a list of messages.
- This PR expands that to track warnings and errors separately, with backward compatibility to the simple "messages" interface. "messages" will be interpreted as either errors or warnings depending on the "is_valid" value.
- Also pulled @iamanandkr 's `get_validation_response` utility into the open source, then modified it to handle errors + warnings, not just messages.

Testing done:
- Ran acasclient unit tests

Other Information:
- This is still missing the ability to store a "Summary" block like the one SEL returns and convey it to the htmlSummary, but I'd prefer to tackle that separately.

### DIFF
--- a/acasclient/__init__.py
+++ b/acasclient/__init__.py
@@ -2,4 +2,4 @@
 
 __author__ = """Brian Bolt"""
 __email__ = 'brian.bolt@boltengineered.com'
-__version__ = '0.1.32'
+__version__ = '0.1.33'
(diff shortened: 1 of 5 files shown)


brianbolt:
LGTM.  

> This is still missing the ability to store a "Summary" block like the one SEL returns and convey it to the htmlSummary, but I'd prefer to tackle that separately.

Yeah, I think a data model where we can separate the view (html) from the data model.  Something to house the data but return html by default or by other formats if asked.

brianbolt (approved):

iamanandkr:
Sorry @bffrost, I totally missed the notification for this. Changes look great to me! I am looking forward to writing validation logics in the BE, everything is so out of the box!

PR #51, opened 2022-04-26
validation: pass commit down to _get_html_summary

bffrost:
## Description
Followup bugfix for recent validation work. Found when I went to use this that the `commit` argument wasn't being passed down properly so it wasn't activating the right summary.

## How Has This Been Tested?
Added new passing unit test that confirms instruction appears when commit=True


brianbolt:
LGTM

brianbolt (approved):

PR #53, opened 2022-05-12
Secure and expose Author and Role management APIs in acasclient

bffrost:
## Description
- Certain scripts or admins need the ability to manage ACAS authors and their roles. However this access is highly sensitive and only admins should be able to create new users or alter their roles via the REST API.
- Added these methods to acasclient so they can be unit tested
- In order to properly test these routes and the admin-vs-non-admin permissions, I ended up adding a bunch of bootstrapping of new users in addition to "bob". The result is that the acasclient test suite can now run on a blank ACAS server that has not had "bob" or the Global project created beforehand.

## How Has This Been Tested?
Confirmed acasclient tests passed on local dev instance.

### DIFF
--- a/acasclient/acasclient.py
+++ b/acasclient/acasclient.py
@@ -247,7 +247,7 @@ class client():
                             data=json.dumps(data),
                             allow_redirects=False)
         resp.raise_for_status()
-        if resp.status_code == 302 or ('location' in resp.headers and resp.headers.location == "/login"):
+        if resp.status_code == 302 and 'location' in resp.headers and resp.headers.get('location') == "/login":
             raise RuntimeError("Failed to login. Please check credentials.")
         return session
 
--- a/acasclient/acasclient.py
+++ b/acasclient/acasclient.py
@@ -244,9 +244,10 @@ class client():
         session = requests.Session()
         resp = session.post("{}/login".format(self.url),
                             headers={'Content-Type': 'application/json'},
-                            data=json.dumps(data))
+                            data=json.dumps(data),
+                            allow_redirects=False)
         resp.raise_for_status()
-        if 'location' in resp.headers and resp.headers.location == "/login":
+        if resp.status_code == 302 or ('location' in resp.headers and resp.headers.location == "/login"):
             raise RuntimeError("Failed to login. Please check credentials.")
         return session
 
--- a/acasclient/acasclient.py
+++ b/acasclient/acasclient.py
@@ -9,6 +9,7 @@ from pathlib import Path
 from pathlib import PurePath
 import re
 import base64
+import hashlib
 from io import StringIO, IOBase
 
 logger = logging.getLogger(__name__)
@@ -1384,3 +1385,85 @@ class client():
                                 .format(self.url, valueId))
         resp.raise_for_status()
         return resp.content
+    
+    def get_authors(self):
+        """
+        Get all authors
+
+        Returns:
+            a list of dict objects representing the authors
+        """
+        resp = self.session.get("{}/api/authors".format(self.url))
+        resp.raise_for_status()
+        return resp.json()
+    
+    def get_author_by_username(self, username):
+        """
+        Get author by username
+            
+        Args:
+            username (str): The username of the author to fetch
+
+        Returns:
+            a dict object representing the author
+        """
+        resp = self.session.get("{}/api/authorByUsername/{}".format(self.url, username))
+        resp.raise_for_status()
+        return resp.json()
+    
+    def create_author(self, author):
+        """
+        Create an author
+
+        Args:
+            author (dict): A dict object representing the author to create
+
+        Returns:
+            a dict object representing the new author
+        """
+        def hash_password(password):
+            """ Calculate the base64-encoded sha1 hash of the password for ACAS built-in authentication """
+            hasher = hashlib.sha1()
+            hasher.update(password.encode('utf-8'))
+            return base64.b64encode(hasher.digest()).decode('utf-8')
+        if 'password' in author:
+            author['password'] = hash_password(author['password'])
+        resp = self.session.post("{}/api/author".format(self.url),
+                                    json=author)
+        resp.raise_for_status()
+        return resp.json()
+    
+    def create_authors(self, authors):
+        """
+        Create authors
+
+        Args:
+            authors (list): A list of dicts representing the authors to create
+            
+        Returns:
+            a list of dict objects representing the saved authors
+        """
+        saved_authors = []
+        for author in authors:
+            saved_authors.append(self.create_author(author))
+        return saved_authors
+    
+    def update_author_roles(self, new_author_roles=None, author_roles_to_delete=None):
+        """
+        Update author roles
+        
+        Args:
+            new_author_roles (list): A list of dicts representing the new author roles to create
+            author_roles_to_delete (list): A list of dicts representing the author roles to delete
+            
+        Returns:
+            a list of dict objects representing the saved author roles
+        """
+        body = {
+            'newAuthorRoles': new_author_roles or [],
+            'authorRolesToDelete': author_roles_to_delete or [],
+        }
+        resp = self.session.post("{}/api/projects/updateProjectRoles".format(self.url),
+                                    json=body)
+        resp.raise_for_status()
+        return resp.json()
--- a/acasclient/acasclient.py
+++ b/acasclient/acasclient.py
@@ -1463,6 +1463,19 @@ class client():
             'newAuthorRoles': new_author_roles or [],
             'authorRolesToDelete': author_roles_to_delete or [],
         }
+        resp = self.session.post("{}/api/updateAuthorRoles".format(self.url),
+                                    json=body)
+        resp.raise_for_status()
+        return resp.json()
+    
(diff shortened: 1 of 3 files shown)


tests/test_lsthing.py
```diff
@@ -77,23 +75,21 @@ def __init__(self, name=None, alias=None, start_date=None, description=None, sta
...
-class TestLsThing(unittest.TestCase):
+class TestLsThing(BaseAcasClientTest):
     """Tests for `acasclient lsthing` package model."""
 
     def setUp(self):
         """Set up test fixtures, if any."""
-        creds = acasclient.get_default_credentials()
-        self.client = acasclient.client(creds)
-        self.tempdir = tempfile.mkdtemp()
+        super().setUp()
```
iamanandkr, line 83: Can we delete it?

acasclient/acasclient.py
```diff
@@ -1383,3 +1385,98 @@ def get_blob_data_by_value_id(self, valueId):
...
+        Args:
+            authors (list): A list of dicts representing the authors to create
+            
+        Returns:
+            a list of dict objects representing the saved authors
+        """
+        saved_authors = []
+        for author in authors:
+            saved_authors.append(self.create_author(author))
+        return saved_authors
```
iamanandkr, line 1449: You may be able to rewrite this as,

```python
return [self.create_author(author) for author in authors]
```

tests/test_acasclient.py
```diff
@@ -222,22 +223,76 @@ def create_thing_with_blob_value(code):
...
+            self.client = acasclient.client(creds)
         self.tempdir = tempfile.mkdtemp()
         # Set TestCase - maxDiff to None to allow for a full diff output when comparing large dictionaries
         self.maxDiff = None
 
     def tearDown(self):
         """Tear down test fixtures, if any."""
         shutil.rmtree(self.tempdir)
+        if self.test_usernames:
+            for username in self.test_usernames:
```
iamanandkr, line 281:
```python
for username in self.test_usernames:
```
should be sufficient. Are we worried about `self.test_usernames` being set to `None`?

tests/test_acasclient.py
```diff
@@ -222,22 +223,76 @@ def create_thing_with_blob_value(code):
...
+                delete_backdoor_user(username)
         self.client.close()
```
iamanandkr, line 283: I believe `self.client.close()` will not be executed if an exception was raised in one of the `delete_backdoor_user` calls. Is this expected?

tests/test_acasclient.py
```diff
@@ -222,22 +223,76 @@ def create_thing_with_blob_value(code):
...
+class TestAcasclient(BaseAcasClientTest):
+    """Tests for `acasclient` package."""
+
+    def setUp(self):
+        """Set up test fixtures, if any."""
+        super().setUp()
+
+    def tearDown(self):
+        """Tear down test fixtures, if any."""
+        super().tearDown()
```
iamanandkr, line 294: Do we need these?

iamanandkr (commented):
The changes look good to me. I have some minor and optional comments. Let me know what you think about those.

acasclient/acasclient.py
```diff
@@ -1383,3 +1385,98 @@ def get_blob_data_by_value_id(self, valueId):
...
         return resp.content
+    
+    def get_authors(self):
+        """
+        Get all authors
+
+        Returns:
+            a list of dict objects representing the authors
+        """
+        resp = self.session.get("{}/api/authors".format(self.url))
```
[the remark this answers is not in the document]
brianbolt: Just noting that when I first created the acasclient package used `format` as f-strings weren't introduced until 3.6 and I was trying to be backwards compatible with 2.7.  I have no issue switching to f-strings and I also don't really have an issue sort of keeping a mix as `format` doesn't appear to be deprecated anytime soon.

acasclient/acasclient.py
```diff
@@ -1383,3 +1385,98 @@ def get_blob_data_by_value_id(self, valueId):
...
+        """
+        Create an author
+
+        Args:
+            author (dict): A dict object representing the author to create
+
+        Returns:
+            a dict object representing the new author
+        """
+        def hash_password(password):
```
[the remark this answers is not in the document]
brianbolt: Just to be a little more specific about @iamanandkr request.  I think if we do create a utils files we should make them named after what they have in them.  I find a general utils.py name does not make the utils functions discoverable which leads to people re-inventing functionality without knowing that similar functions already exist.

acasclient/acasclient.py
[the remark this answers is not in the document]
brianbolt: One more opinion. I do find f-strings to be a little more readable so they are my preference now

tests/test_acasclient.py
```diff
@@ -22,6 +21,8 @@
 M  END
 """
 
+ACAS_NODEAPI_BASE_URL = "http://localhost:3001"
```
brianbolt, line 24: Trying to think of a way to communicate a message when a someone doesn't have access to port 3001.  Wonder if a decorator on these routes that use this would work.

brianbolt (commented):
I had one comment but @iamanandkr has a lot of good comments so I'll let you iterate.

acasclient/acasclient.py
```diff
@@ -1383,3 +1385,98 @@ def get_blob_data_by_value_id(self, valueId):
...
+            """ Calculate the base64-encoded sha1 hash of the password for ACAS built-in authentication """
+            hasher = hashlib.sha1()
+            hasher.update(password.encode('utf-8'))
+            return base64.b64encode(hasher.digest()).decode('utf-8')
+        if 'password' in author:
```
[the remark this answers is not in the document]
bffrost: I appreciate this comment and the thought it provokes. Right now I don't think it's worth defining these as constants, since I don't see a lot of repetition of strings _across_ different functions. If we later migrate more of this towards object-oriented classes, that will be a good time to define these constants in a single place.

acasclient/acasclient.py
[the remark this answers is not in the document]
bffrost: I'm going to pass on the utils file idea simply because this utility function isn't used anywhere else. I purposely defined it as an in-line function since it is only used in this one function.

I'll apply the type hints suggestion though. I really like those too.

tests/test_acasclient.py
bffrost: Good point, wrapping this in a try/finally block

tests/test_acasclient.py
bffrost: No, I supposed not. i had left them in as placeholders if we wanted to extend the setup & cleanup but we can always add them back in if we need to do that.

tests/test_acasclient.py
bffrost: Good suggestion. I've implemented this as a decorator and done some testing by temporarily tweaking the ACAS_NODEAPI_BASE_URL to 30001 to simulate inaccessibility. As far as I can tell, the tests will run without port 3001 as long as bob and Global have already been created, and otherwise they will all be skipped. And my last test which explicitly requires the node api to create a test user will always be skipped unless node api is available.

acasclient/acasclient.py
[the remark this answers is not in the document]
bffrost: I'm happy to switch over to f-strings but I'm going to handle that in a separate PR, along with type hints.

acasclient/acasclient.py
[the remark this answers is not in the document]
bffrost: Decided to do the type hints in a separate PR since it will be a lot of changes. I'll do it along with f-strings.

tests/test_acasclient.py
bffrost: Yep, you're right. Adopted this.

acasclient/acasclient.py
```diff
@@ -244,9 +245,10 @@ def getSession(self):
         session = requests.Session()
         resp = session.post("{}/login".format(self.url),
                             headers={'Content-Type': 'application/json'},
-                            data=json.dumps(data))
+                            data=json.dumps(data),
+                            allow_redirects=False)
         resp.raise_for_status()
-        if 'location' in resp.headers and resp.headers.location == "/login":
+        if resp.status_code == 302 and 'location' in resp.headers and resp.headers.get('location') == "/login":
```
brianbolt, line 251: Nit: Just to future proof this a little in case we end up with a /acas prefix to our urls, it might be good to set this to:
```
resp.headers.get('location').endswith('/login')
```

That being said, we should probably be giving a 401 when hitting login via api so that probably a better fix down the line.

tests/test_acasclient.py
brianbolt: Nice!

brianbolt (approved):
LGTM.  Thanks for all the work on this.

iamanandkr (approved):
LGTM!

acasclient/acasclient.py
bffrost: Yeah, going to leave this as-is. I'm not sure how it'd handle things like `/login?redirect=foo` and I agree a status code would be a better fix, but considering that out of scope.

PR #57, opened 2022-05-19
Dose response curve validation tests

brianbolt:
## Description
Adds for curve validation features.
  - Added a validation file which covers many of the cases used in curve fit validation
  - Wrote tests to make sure curve fit validation messages match expected messages
  - Added parameter`validate_dose_response_curves` parameter to `experiment_loader` to turn off curve validation (to mimic UI behavior which turns off curve validation when being run from dose response module.
  - Re-arranged experiment loader tests to their own section.  This messed up the number ordering of tests but I didn't fix that here. 
 
Screen Shot 2022-05-18 at 4 57 20 PM

*More features folded into this PR*

 - Decorators
   - Ran into an issue where I needed to know the state of the DB before tests.  i.e. that the basic cmpd reg file was loaded before the test.  We were able to avoid knowing this before some of the tests, mainly because of the order in which were were executing tests worked out.  i.e. Some tests didn't know that CMPD-000001 was loaded before the test but because a previous test had loaded it, it was ok.  To be a bit more unit test driven, without having to reorganize all the tests into specific classes which know the state, I added a few decorators require basic cmpdreg load was present or absent in the database and also one that confirms that the basic experiment load was done in the database.
 - tearDownClass and setUpClass vs setUp and tearDown
   -   I found that the setUp and tearDown functions were getting run for each and every test, which was unnecessary for most things so I added "setUpClass" and "tearDownClass" which only get run once per class instantiation.  I kept the tempdir setUp and tearDown as that is fast and potentially important to clear between tests.
   - Additionally to the tearDownClass, I also added functionality to delete all experiments, cmpdreg bulk loads (in that order) and also all projects created.  This doesn't complete remove everything from the database but it does keep things somewhat clear when re-running tests.

## Related Issue
ACAS-273

## How Has This Been Tested?
Ran ACAS client tests!


tests/test_acasclient.py
```diff
@@ -1797,4 +1595,299 @@ def test_044_author_and_role_apis(self):
         with self.assertRaises(requests.HTTPError) as context:
             acas_user_author_role['userName'] = test_username
             user_client.update_author_roles(author_roles_to_delete=[acas_user_author_role])
-        self.assertIn('500 Server Error', str(context.exception))
\ No newline at end of file
+        self.assertIn('500 Server Error', str(context.exception))
+
+
+
+def experiment_load_test(data_file_to_upload, dry_run_mode, self):
```
bffrost, line 1602: This function uses `self` strangely. Can we move this function into TestExperimentLoader or BaseAcasClientTest and move self to the beginning so it is called like `self.experiment_load_test(data_file_to_upload, True)`? That would be more pythonic

tests/test_acasclient.py
```diff
@@ -1797,4 +1595,299 @@ def test_044_author_and_role_apis(self):
...
+    file_name = file_name.replace(".csv", ".txt")
+    temp_file_path = Path(self.tempdir, file_name)
+    # Change the delim to the new delim
+    with open(data_file_to_upload, 'r') as f:
+        with open(temp_file_path, 'w') as f2:
+            for line in f:
+                f2.write(line.replace(',', "\t"))
+    return temp_file_path
+
+class TestExperimentLoader(BaseAcasClientTest):
```
bffrost, line 1629: I like splitting these out into a separate test, but don't they require some compound loading to happen first?
It would be better to have some `setUp` logic that confirms the necessary compound lots exist, and registers them if not, that way this `TestExperimentLoader` test case can be run standalone.

bffrost (changes requested):
I like the refactor. Left a few comments on things I think you should change

tests/test_acasclient.py
```diff
@@ -1797,4 +1595,299 @@ def test_044_author_and_role_apis(self):
...
+    self.assertIn('hasError', response)
+    self.assertIn('hasWarning', response)
+    self.assertIn('transactionId', response)
+    if dry_run_mode:
+        self.assertIsNone(response['transactionId'])
+    else:
+        self.assertIsNotNone(response['transactionId'])
+    return response
+
+def csv_to_txt(data_file_to_upload, self):
```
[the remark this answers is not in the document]
brianbolt: It was kind of but I refactored it to pass a more specific variable.

tests/test_acasclient.py
brianbolt: Refactored, thanks.

tests/test_acasclient.py
brianbolt: Did major refactor to fix this.  Added setUpClass and tearDownClass as well as a few decorators for ensuring the database was in the proper state for the test.

tests/test_acasclient.py
```diff
@@ -1797,4 +1770,296 @@ def test_044_author_and_role_apis(self):
...
+            # about 9 seconds. This is a sanity check to make sure the
+            # dry run hasn't slowed significantly.
+            with Timeout(seconds=25):
+                data_file_to_upload = Path(__file__).resolve()\
+                    .parent.joinpath('test_acasclient', '50k-lines.csv')
+                self.experiment_load_test(data_file_to_upload, True)
+        except TimeoutError:
+            self.fail("Timeout error")
+        else:
+            pass
```
iamanandkr, line 1874: Can be removed.

tests/test_acasclient.py
```diff
@@ -1797,4 +1770,296 @@ def test_044_author_and_role_apis(self):
...
+        # Test dose response curve validation
+        data_file_to_upload = Path(__file__).resolve()\
+            .parent.joinpath('test_acasclient', '4 parameter D-R-validation.csv')
+
+        response = self.experiment_load_test(data_file_to_upload, True)
+
+        # When loading Dose Resposne format but not having ACAS fit the curves, we shold get a dose response summary table
+        self.assertTrue(response['results']['htmlSummary'].find("bv_doseResponseSummaryTable") != -1)
+
+        # print(json.dumps(response['errorMessages'], sort_keys=True, indent=4))
```
iamanandkr, line 1887: To be removed?

tests/test_acasclient.py
```diff
@@ -267,9 +267,65 @@ def get_or_create_global_project():
...
+        experiments = self.client.get_experiment_by_name("BLAH")
+        if len(experiments) == 0:
+            data_file_to_upload = Path(__file__).resolve()\
+                            .parent.joinpath('test_acasclient', 'uniform-commas-with-quoted-text.csv')
+            response = self.client.\
+                experiment_loader(data_file_to_upload, "bob", False)
+            experiments = self.client.get_experiment_by_name("BLAH")
+        for experiment in experiments:
+            if experiment['ignored'] == False and experiment['deleted'] == False:
+                break;
```
iamanandkr, line 310: `;` not needed.

iamanandkr (commented):
LGTM! I have some minor comments/suggestions.

tests/test_acasclient.py
```diff
@@ -291,19 +347,286 @@ def setUp(self):
...
+            raise Exception("Failed to delete all experiments: " + str(not_deleted_experiments))
+                    
+
+    def delete_all_cmpd_reg_bulk_load_files(self):
+        """ Deletes all cmpdreg bulk load files in order by id """
+
+        files = self.client.\
+            get_cmpdreg_bulk_load_files()
+        
+        # sort by id
```
bffrost, line 429: This bit is interesting. I would think this code would sort the IDs ascending and then try to delete in the same order the files were loaded.
Wouldn't we want to delete the files in the opposite order from how they were loaded (i.e. remove most recent to prevent there being dependencies)? Am I missing something here?

tests/test_acasclient.py
```diff
@@ -291,19 +347,286 @@ def setUp(self):
...
+            get_cmpdreg_bulk_load_files()
+        if len(files) > 0:
+            # Get the ids of all the files
+            ids = [file['id'] for file in files]
+            # Throw exception not failure
+            raise ValueError(f"Failed to delete some cmpd reg bulk load files: {ids}")
+
+
+    def delete_all_projects(self):
+        """ Deletes all cmpdreg bulk load files in order by id """
```
bffrost, line 445: nit: copy/pasted docstring should mention projects

bffrost (commented):
This is looking good. I like the refactor and the new wrappers.
Noted a few small things. Once you address Anand's and my comments I think you'll be good to go.

tests/test_acasclient.py
```diff
@@ -267,9 +267,65 @@ def get_or_create_global_project():
     output = r.json()
     return output["messages"]
 
+def requires_basic_cmpd_reg_load(func):
+    """
+    Decorator to load the basic cmpdreg data if it is not already loaded
+    """
+    @wraps(func)
```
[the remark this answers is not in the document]
brianbolt: Yeah, I'm not sure best practice.  I knew I wnted to use a decorator and I also found this to be the hardest part of this change mainly because I originally tried to modify self to add some meta about whether the experiment was loaded or not currently (to track the state without having to query the db), but it was failing so I fell back to querying the DB which I think is actually a more robust solution to this.

tests/test_acasclient.py
brianbolt: Too much javascript. Thanks.

tests/test_acasclient.py
```diff
@@ -267,9 +267,65 @@ def get_or_create_global_project():
...
+        if len(experiments) == 0:
+            data_file_to_upload = Path(__file__).resolve()\
+                            .parent.joinpath('test_acasclient', 'uniform-commas-with-quoted-text.csv')
+            response = self.client.\
+                experiment_loader(data_file_to_upload, "bob", False)
+            experiments = self.client.get_experiment_by_name("BLAH")
+        for experiment in experiments:
+            if experiment['ignored'] == False and experiment['deleted'] == False:
+                break;
+        return func(self, experiment)
```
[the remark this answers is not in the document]
brianbolt: Yes, thanks, I made that more clear and added a default of None.

tests/test_acasclient.py
brianbolt: I left that there on purpose in case we need to update the expected results dict.  I added a new comment to clarify why the comment was left in the code.

PR #60, opened 2022-06-06
Add back tear down of cmpd reg bulk load files, fix test045_register_…

brianbolt:
## Description
I had inadvertently removed a key tearDown of compd reg bulk load files before merging ACAS-273
This also fixes `test045_register_sdf_case_insensitive` which expected cmpds to be loaded, I modified the test to include decorator `@requires_absent_basic_cmpd_reg_load` and fixed a test to expect new compounds

## Related Issue
ACAS-273

## How Has This Been Tested?
ACAS Client tests from empty db.


bffrost (approved):
LGTM

PR #65, opened 2022-06-21
Dale Adding Tests for  ACAS-99 and ACAS-40

dalejerikson:
Adding tests for ACAS-99 and ACAS-40

## Description
Adding tests to test_acasclient.py

test_047_load_sdf_with_salts - ACAS should throw error when salts found in both structure and sdf 

test_048_warn_existing_compound_new_id - ACAS should throw warnings if molecule registered gets assigned a new parent if stereo category or stereo comment is different than existing parent  

## How Has This Been Tested?
By running the tests locally :)


tests/test_acasclient.py
```diff
@@ -2296,6 +2296,144 @@ def test046_cmpdreg_admin_crud(self):
         self.client.delete_cmpdreg_vendor(vendor['id'])
         
 
+    def test_047_load_sdf_with_salts(self):
+        """
+        Test to Load SDF File w/ Salts to Determine Process
```
bffrost, line 2301: nit: I think you can update this docstring now that you understand what this is testing

tests/test_acasclient.py
```diff
@@ -2296,6 +2296,144 @@ def test046_cmpdreg_admin_crud(self):
...
+                "sdfProperty": "Lot Salt Name"
+            },
+            {
+                "dbProperty": "Lot Salt Equivalents",
+                "defaultVal": None,
+                "required": False,
+                "sdfProperty": "Lot Salt Equivalents"
+            }
+        ]
+
```
bffrost, line 2343: nit: This could do with a code comment indicating that you're ensuring `HCl` is registered as a salt since it's needed for your test.

tests/test_acasclient.py
```diff
@@ -2296,6 +2296,144 @@ def test046_cmpdreg_admin_crud(self):
...
+        # Get Salt Abbrevs. Treat salts separately since they are not a standard codetable
+        salts = self.client.get_salts()
+        # Create Salt Abbrev
+        if SALT_ABBREV.lower() not in [s['abbrev'].lower() for s in salts]:
+            salt = self.client.create_salt(abbrev=SALT_ABBREV, name=SALT_ABBREV, mol_structure=SALT_MOL)
+            self.assertIsNotNone(salt.get('id'))
+
+        response = self.client.register_sdf(test_047_load_sdf_with_salts_file, "bob",
+                                            mappings)
+        print(response)
```
bffrost, line 2355: nit: might not need this `print` anymore

tests/test_acasclient.py
```diff
@@ -2296,6 +2296,144 @@ def test046_cmpdreg_admin_crud(self):
...
+        self.assertIn('report_files', response)
+        self.assertIn('summary', response)
```
bffrost, line 2357: I was expecting a few more checks here. What about checking that the error is thrown x number of times and that the records you expect to be successful were successfully registered?

tests/test_acasclient.py
```diff
@@ -2296,6 +2296,144 @@ def test046_cmpdreg_admin_crud(self):
...
+            {
+                "dbProperty": "Parent Stereo Category",
+                "defaultVal": "mixture",
+                "required": True,
+                "sdfProperty": None
+            },
+        ]
+        response = self.client.register_sdf(test_048_warn_existing_compound_new_id_file_one, "bob",
+                                            mappings)
+        print(response)
```
bffrost, line 2397: I think you can remove the `print` and you should add some more detailed assertions to check that no error is thrown here.

bffrost (changes requested):
The content and flow of these tests looks great, but you should add some more checks to assert that the errors you expect are actually getting thrown. You'll need to parse the response to get those error messages out.

tests/test_acasclient.py
dalejerikson: Implemented

tests/test_acasclient.py
dalejerikson: Implemented

bffrost (approved):
Excellent, looks great!

PR #74, opened 2022-07-23
ACAS-343: Run acasclient tests against appropriate version of ACAS (PR for release/2022.1.x)

bffrost:
## Description
Modify acasclient GitHub Actions to run against correlated ACAS branch. This is part of a series of 4 PRs to update both release branches of both the `acasclient` and `acas` repos.
- For commits to `main` and `release/*` branches (i.e. `release/1.13.7`, `release/2022.1.x`), run tests against the corresponding ACAS branch by the same name.
- For pull requests, run acasclient tests against the target branch, i.e. ACAS `release/2022.1.x` if the acasclient PR is targeting `release/2022.1.x`

## How Has This Been Tested?
The GitHub action passed and I confirmed it ran `test_004_delete_lot` which is a newer, 2022.1.x only test.


.github/workflows/run_tests.yml
```diff
@@ -23,15 +23,31 @@ jobs:
...
+      - name: Set ACAS_REF to the PR destination branch ${{ github.base_ref }}
+        run: |
+          echo "ACAS_REF=${{ github.base_ref }}" >> $GITHUB_ENV
+        if: github.event_name == 'pull_request'
+      - name: Set ACAS_TAG to the docker escaped version of ${{ env.ACAS_REF }}
+        run: |
+          echo "ACAS_TAG=$(echo ${{ env.ACAS_REF }} | sed 's/\//-/g')" >> $GITHUB_ENV
+      - name: Override ACAS_REF and ACAS_TAG if branch is "main"
+        run: |
+          echo "ACAS_REF=master" >> $GITHUB_ENV
```
brianbolt, line 42: Instead of hardcoding master here you could do:

```
curl -s https://api.github.com/repos/mcneilco/acas | jq -r '.default_branch'
```

brianbolt (approved):
LGTM.  Approving but made a easy suggestion to get the default branch of acas from the api instead of hardcoding master.

.github/workflows/run_tests.yml
```diff
@@ -2,7 +2,7 @@ name: Launch ACAS and run python tests
 
 on:
   push:
-    branches: ["main"]
+    branches: ["main", "release/*"]
```
brianbolt, line 5: I just found a nice feature which is to use: $default-branch macro:
Example:
https://github.com/actions/starter-workflows/blob/main/ci/docker-image.yml#L5

https://github.blog/changelog/2020-07-22-github-actions-better-support-for-alternative-default-branch-names/

brianbolt (approved):
My second review is still LGTM. But I think we should add this $default-branch macro (see my comments)

.github/workflows/run_tests.yml
brianbolt: Ignore this comment.  I guess the $default-branch macro is for action templates, not for the actual GitHub actions yaml:

https://github.com/github/docs/discussions/5071

brianbolt (changes requested):
Sorry, I keep changing my mind on this.  Because these changes are also occurring on 1.13.7 branch (which is the downstream release). I think we need to up merge those commits into release/2022.1.x rather than making these new commits.

.github/workflows/run_tests.yml
bffrost: Hm.... do we actually want the default branch though? I think we already cover testing against the latest release branch via the corresponding release branch of acasclient.

My unspoken assumption here is that we'd upmerge `release/1.13.7` -> `release/2022.1.x` -> `master` in acas, and `release/1.13.7` -> `release/2022.1.x` -> `main` in acasclient, so I was hardcoding a correspondence between `master` and `main`

bffrost:
Makes sense. I'm closing this PR in favor of #76 which goes into `release/1.13.7`. Then once that's merged I'll go ahead with the upmerging you describe.

PR #75, opened 2022-07-25
Fast forward release/1.13.7 to v0.1.35 tag

bffrost:
## Description
I determined from some local testing and looking at the git log that this later commit, `v0.1.35` is safe and compatible with `release/1.13.7`. My breaking change went in after this tag.

This PR contains no new commits, just advancing the release branch to a more recent point in history.

## How Has This Been Tested?
Ran acasclient tests locally, determined causes of failures were trivial / not related to version incompatibility.
I will file a follow-up PR with fixes for flaky tests.


brianbolt (approved):
LGTM

PR #76, opened 2022-07-26
ACAS-343: Run acasclient tests against appropriate version of ACAS (PR for release/1.13.7)

bffrost:
## Description
Modify acasclient GitHub Actions to run against correlated ACAS branch. This is the 2nd of 4 PRs, following #74 . This PR applies the same change to `release/1.13.7`.

Changes:
- For commits to main and release/* branches (i.e. release/1.13.7, release/2022.1.x), run tests against the corresponding ACAS branch by the same name.
- For pull requests, run acasclient tests against the target branch, i.e. ACAS release/2022.1.x if the acasclient PR is targeting release/2022.1.x
- Fixed a few flaky / failing tests specific to `release/1.13.7` branch

## How Has This Been Tested?
See successful tests job in the "check" on this PR: https://github.com/mcneilco/acasclient/runs/7511431603?check_suite_focus=true

PR #79, opened 2022-07-27
add sed for current branch case

bffrost:
## Description
Follow-up to #76 which failed after merging. See https://github.com/mcneilco/acasclient/actions/runs/2750006213

## How Has This Been Tested?
Will test by merging this.


bffrost:
targeted wrong base branch. going to close and try again.

PR #81, opened 2022-07-28
ACAS-367: partial revert

bffrost:
## Description
Original fix for ACAS-367 tried to fix two issues - one with misaligned quote escaping between R and Roo, and another with R being unable to read files with `""` escaped double quotes, rather than `\"` escaped.
The fix for the second issue introduced new bugs, so it has now been reverted: https://github.com/mcneilco/racas/commit/952d5a97d5dcae94b8215c21688dc5ddbda63c46

This PR updates the python test for that second issue, and switches it to expect failure. I thought that keeping the test and recording the fact that it fails was more valuable than simply removing the test, but I'm open to other opinions there.

## How Has This Been Tested?
Tested acasclient locally. Will confirm with automated test on this PR.


brianbolt (approved):
LGTM

PR #82, opened 2022-08-09
Add 401 status code check and test for failed login to client initialization

brianbolt:
## Description

- For ACAS 2021.1.x detect no login by checking 401 status code.
- Keeps backwards compatibility for 302 and /login redirect for older versions of ACAS.

## Related Issue
ACAS-389

## How Has This Been Tested?
Verified that new and old ACAS work with this detection method.


bffrost (approved):
LGTM

PR #84, opened 2022-08-18
Expanded tests for delete_lot, including coverage for disableDeleteMyLots config

brianbolt:
## Description
 Modify delete_lot tests to account for new default config `client.cmpdreg.metaLot.disableDeleteMyLots` which disables all deleting of lots if user is not a creg admin

## Related Issue
ACAS-408

## How Has This Been Tested?
Ran default acas client tests with default acas setting `client.cmpdreg.metaLot.disableDeleteMyLots` = `true`
Ran `TODO` tests which are the non-default tests for when `client.cmpdreg.metaLot.disableDeleteMyLots` = `false`

### DIFF
--- a/tests/test_acasclient.py
+++ b/tests/test_acasclient.py
@@ -400,10 +400,10 @@ class BaseAcasClientTest(unittest.TestCase):
             self.client.close()
 
     @requires_node_api
-    def create_and_connect_backdoor_user(self, username = None, password = None, **kwargs):
+    def create_and_connect_backdoor_user(self, username = None, password = None, prefix = "acas-user-", **kwargs):
         """ Creates a backdoor user and connects them to the ACAS node API """
         if username is None:
-            username = "acas-user-"+str(uuid.uuid4())
+            username = prefix+str(uuid.uuid4())
         if password is None:
             password = str(uuid.uuid4())
         create_backdoor_user(username = username, password = password, **kwargs)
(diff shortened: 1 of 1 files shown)


tests/test_acasclient.py
```diff
@@ -2719,56 +2721,115 @@ def can_delete_lot(self, user_client, lot_corp_name, set_owner_first=True):
             self.assertTrue(response['success'])
             return True
 
-        # Global compound, with experiment in Global project
-        ## Deny Rule: Not an acas user so can't delete because dependent experiment exists
+        def check_lot_exists_in_experiment(self, lot_corp_name, experiment_code_name):
+            experiment = self.client.get_experiment_by_code(experiment_code_name, full = True)
+            # when experiment is deleted, the analysis groups for the lot are deleted
+            for analysis_group in experiment["analysisGroups"]:
+                if analysis_group["deleted"] == True:
```
bffrost, line 2728: Should we be checking `ignored` here too?

tests/test_acasclient.py
```diff
@@ -2719,56 +2721,115 @@ def can_delete_lot(self, user_client, lot_corp_name, set_owner_first=True):
...
+            experiment_loader(file_to_upload, "bob", False)
+        self.assertTrue(check_lot_exists_in_experiment(self, restricted_lot_corp_name, response['results']['experimentCode']))
+                        
+        # Deny rule: Creg admin is not an acas user
+        self.assertFalse(can_delete_lot(self, cmpdreg_admin, restricted_lot_corp_name, set_owner_first=True))
+ 
+        # Deny rule: ACAS admin is not a creg admin
+        self.assertFalse(can_delete_lot(self, acas_admin, restricted_lot_corp_name, set_owner_first=True))
+
+        # Deny rule: CregAdmin/ACASAdmin can delete the lot and assay data
```
bffrost, line 2772: This is an Allow rule, right?

bffrost (changes requested):
Noted some super small things to fix, but otherwise makes sense. I understand why you've left in that commented out portion for once we develop a method to test under different config scenarios.

tests/test_acasclient.py
brianbolt: Fixed thanks.

tests/test_acasclient.py
brianbolt: Fixed thanks.

brianbolt:
I snuck in one additional fix to allow for optional username prefix to the create_and_connect_backdoor_user function.

bffrost (approved):
LGTM

PR #87, opened 2022-08-28
Add tests for checking modified date on updating Lots.

iamanandkr:
Add tests for checking `modifiedDate` on updating `Lot`.

## Description

When a `Lot` is edited, the `modifiedDate` field is not updated. Add tests for checking the `modifiedDate` on updating a lot.

## Related Issue

[ACAS-364](https://jira.schrodinger.com/browse/ACAS-364).

## How Has This Been Tested?

TDD.

### DIFF
--- a/tests/test_acasclient.py
+++ b/tests/test_acasclient.py
@@ -2642,6 +2642,17 @@ class TestCmpdReg(BaseAcasClientTest):
         self.assertEqual(meta_lot['lot']['corpName'], restricted_project_lot_corp_name)
         self.assertIsNone(meta_lot["lot"]["modifiedDate"])  # Lot is not modified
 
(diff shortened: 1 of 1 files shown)


dalejerikson (commented):
Seeking others opinion on this...

Is it reasonable to check that the modifiedDate is not none or should it be a step further where we obtain the current date via Python's `datetime` and see if the month and day match the `modifiedDate` in the response?

bffrost:
@dalejerikson Your point is a valid one, but given how tricky dates can be, and that these CI tests may run at any time of day, I'd rather not open ourselves up to potential flaky test issues around dates that might occur if this were running right at midnight of whatever timezone. IMO it's enough to check that modifiedDate is not None.

PR #88, opened 2022-08-29
ACAS-404: Update swap_parent_structures API to return error information.

iamanandkr:
Update `swap_parent_structures` API to instead return a dictionary with `hasError` and `errorMessage` keys.

## Description

`swap_parent_structures` API doesn't produce any meaningful error message in case of failure. Here we update the API to pass the json with `hasError` & `errorMessage` returned from the roo layer to the caller.

## Related Issue

[ACAS-404](https://jira.schrodinger.com/browse/ACAS-404).

## How Has This Been Tested?

TDD.

### DIFF
--- a/acasclient/acasclient.py
+++ b/acasclient/acasclient.py
@@ -11,6 +11,8 @@ import re
 import base64
 import hashlib
 from io import StringIO, IOBase
+from typing import Dict
+
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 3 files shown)


dalejerikson (approved):
LGTM

bffrost (approved):
LGTM. I did notice that the CI test had failed, so I've triggered a re-run. If it fails for the same reason again we should investigate.

PR #96, opened 2022-09-29
ACAS-464: cast list to string since that's what backend expects

bffrost:
## Description
Was seeing an issue with searching using corpNameList, and traced it down to the Roo backend expecting a comma-separated string rather than an actual list.
Put in a patch to acasclient to force list input to string.

## How Has This Been Tested?
Reproduced issue with acasclient test, and confirmed it was fixed after the change


brianbolt (approved):
LGTM

PR #97, opened 2022-09-29
Show `swapParentStructures` errors to the client.

iamanandkr:
## Description

`swapParentStructures` API in `acas-roo-server` is updated to return a `BAD_REQUEST` when the swap of parent structures fails due to invalid corporate name or other invalid criteria. Here we update the client code to ignore `400` status code in response as we want the error message to be propagated to the end user.

## Related Issue

[ACAS-430](https://jira.schrodinger.com/browse/ACAS-430)
https://github.com/mcneilco/acas-roo-server/pull/369
## How Has This Been Tested?

Ran all acasclient tests.


bffrost (approved):
LGTM

PR #100, opened 2022-10-07
ACAS-271: add tests for originalFileName

bffrost:
## Description
I should've added these on my earlier PR, but now I have. Added a few checks to confirm the originalFileName is being passed and preserved through SDF registration.

## How Has This Been Tested?
Confirmed all tests passed.


hhan-schrodinger (approved):
Looks good to me!

PR #101, opened 2022-11-10
Test for Correct Updating of Parent Aliases Through Bulk Loader 

dalejerikson:
This addition is to pair with the fix for the following bug report...

**Steps to Reproduce**

Bulk load a new structure with a Parent Alias (or list of several) mapped in. This creates a new parent with the alias(es), and a new lot.

Then load the same structure with a different Parent Alias value (or list of several). This will create a new lot on the existing parent.

**Previous Bug**

A new lot is created, but the alias(es) from the second upload is ignored. The parent ends up with only the alias(es) from the first upload.

**Expected Results**

A new lot will be created and the existing parent will be updated to add the new Parent Alias value(s). The parent will end up with the aliases from both uploads.

### DIFF
--- a/tests/test_acasclient.py
+++ b/tests/test_acasclient.py
@@ -2349,7 +2349,7 @@ class TestAcasclient(BaseAcasClientTest):
         self.assertIn('Number of entries processed: 1000', response['summary'])
         self.assertIn('Number of entries with error: 1', response['summary'])
 
-    def test_50_delete_ls_thing(self):
+    def test_050_delete_ls_thing(self):
         code = str(uuid.uuid4())
         ls_thing = create_project_thing(code)
         saved_ls_thing = self.client.save_ls_thing(ls_thing)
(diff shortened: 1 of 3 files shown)


dalejerikson:
FIx PR in Roo: https://github.com/mcneilco/acas-roo-server/pull/391

tests/test_acasclient.py
```diff
@@ -2367,6 +2367,203 @@ def test_50_delete_ls_thing(self):
...
+        corp_name = "CMPD-0000051"
+        alias_one = "First Alias"
+        alias_two = "Second Alias"
+        alias_three = "Third Alias"
+        alias_four = "Fourth Alias"
+        alias_five = "Fifth Alias"
+
+        # Get aliases
+        aliases = self.client.get_parent_aliases(corp_name)
+        self.assertEqual(len(aliases), 3)
```
bffrost, line 2541: @dalejerikson  Could you add a test case where we load the same file a second time, and confirm we still only get 3 aliases? Just to test that the string comparison logic (which i see you fixed on the acas-roo-server PR) is working as expected.

bffrost (changes requested):
Requesting one more case to be added to this test, but otherwise looks good!

bffrost:
@dalejerikson I also noticed your self-reminder on the Jira case:
Self Reminder: Write test where existing parent has 3 (or so) aliases and the incoming parent from file has 2 (or so) and there is some overlap -> ensure that all aliases are captured (and no duplicates) and that the modified by and modified date fields are updated as well -> acas client 

that'd be a good case to add as well.

tests/test_acasclient.py
dalejerikson: Will do!

dalejerikson: Completed

dalejerikson:
> @dalejerikson I also noticed your self-reminder on the Jira case: Self Reminder: Write test where existing parent has 3 (or so) aliases and the incoming parent from file has 2 (or so) and there is some overlap -> ensure that all aliases are captured (and no duplicates) and that the modified by and modified date fields are updated as well -> acas client
> 
> that'd be a good case to add as well.

Added this case as well. However, looks like there's not a history in test_acasclient for testing modified by and last modified by dates. Should this be added or is this sort of testing usually done manually?

dalejerikson:
Looks like GitHub Actions run failed. When comparing it to my local it seems that the `client.cmpdreg.metaLot.allowDuplicateParentAliases` is flipped. Interesting. Will need to investigate where this is happening in roo. In the past, I would expect this to have worked without the intervention of my changes.

bffrost:
@dalejerikson I think the GitHub Actions run is happening on a build of ACAS that doesn't include your Roo changes, since your Roo PR https://github.com/mcneilco/acas-roo-server/pull/391 hasn't been merged. If your tests are passing locally then I think they're fine. Please focus on merging https://github.com/mcneilco/acas-roo-server/pull/391 then come back and re-run the GitHub Actions build on this PR after your Roo changes are merged.

tests/test_acasclient.py
```diff
@@ -2544,6 +2546,24 @@ def test_051_bulk_load_update_parent_alias(self):
         self.assertIn(alias_two, aliases)
         self.assertIn(alias_three, aliases) 
 
+        # Redo the same file (copy) to see if the result is the same 
+        # and aliases aren't just being appended without comparison 
+        test_051_upload_file_file_one_copy = Path(__file__).resolve().parent\
```
bffrost, line 2551: Is this literally the same file twice with different filenames? Why not just upload `test_051_register_parent_aliases.sdf` again?

tests/test_acasclient.py
```diff
@@ -2544,6 +2546,24 @@ def test_051_bulk_load_update_parent_alias(self):
...
+            .joinpath('test_acasclient', 'test_051_register_parent_aliases_copy.sdf')
+        files = self.client.upload_files([test_051_upload_file_file_one_copy])
+        request["fileName"] = files['files'][0]["name"]
+        response = self.client.register_sdf_request(request)
+        self.assertIn('reportFiles', response[0])
+        self.assertIn('summary', response[0])
+        self.assertIn('Number of entries processed', response[0]['summary'])
```
bffrost, line 2558: I know it's not your primary purpose in this PR, but could you add a check for how many compounds vs lots of new compounds were registered? This could function as a test of registering subsequent lots, which is something we're missing and would've helped me catch the bug I recently introduced.

bffrost (changes requested):
Nearly there, tests look good. One question about the copied file, and a request to add a few more checks so this test can do double duty

tests/test_acasclient.py
dalejerikson: Yep, I'll add that!

tests/test_acasclient.py
dalejerikson: It is. I thought it was reasonable to replicate the circumstance where someone is trying to register the same compound but some of the file metadata information (e.g. data created, date modified, etc.) is slightly different. Not a big deal. I can re-register the same file.

bffrost: Ah, okay. Not a big deal. I was just wondering why. I don't think we track any of the file metadata information, but it's fine for you to keep this as-is. Consider this one truly optional.

dalejerikson:
Note: GitHub Actions will fail running this (even after associated Roo merge) due to separate bug of "Duplicate Parent Alias Error Thrown on Same Parent" where "There is an error where ACAS complains that parents cannot have duplicate aliases in the system (NonUniqueAliasException in Roo)" 

**Reproduction Steps** 

Note: The configuration of ACAS needs to be set to not allow duplicates in parent aliases 

1. Utilize bulk loader to validate the sdf of a compound with a single or set of aliases 
2. Register compound 
3. Utilize bulk loader (again) to validate a different sdf of the same compound where the only difference is the alias(es) it has. At least one alias should present in the previous successful compound registration 
4. Attempted to register the compound in the second file

bffrost:
@dalejerikson Thanks for the steps to reproduce, and for filing a ticket for that bug. However I'm still confused about why that bug would cause GH actions failures here.
Our default configs _allow_ duplicate aliases: https://github.com/mcneilco/acas/blob/release/2022.4.x/conf/config.properties.example#L708 so why would this GH actions test be failing due to a duplicate alias error?

bffrost:
@dalejerikson I figured out a fix for the CI issue - I'm pretty sure it was one of the previously tests (my large SDF loading test) not cleaning up after themselves properly. I've added a fix: https://github.com/mcneilco/acasclient/pull/101/commits/7bc46f34db4e4a94b939339778b3f2a4fb57b1bf

Once that is in place, I was reliably getting this failure:
```
  File "/Users/frost/Projects/ACAS/acasclient/tests/test_acasclient.py", line 2599, in test_051_bulk_load_update_parent_alias
    self.assertEqual(len(aliases), 7)
AssertionError: 9 != 7
```

I did a tiny bit of debugging and found that the extra aliases had a leading space in them:
Screen Shot 2022-11-21 at 11 52 13 AM

This is coming from `test_051_register_parent_aliases_overlap_aliases.sdf` which has:
```
>  <Parent Alias>
Fourth Alias; Fifth Alias; Sixth Alias; Seventh Alias;
```

Ultimately I think your test is fine, but it's uncovered a bug in ACAS which is that we aren't stripping leading / trailing whitespace from aliases when we read in multiple semicolon-delimited values. I'm going to explore a fix.

tests/test_acasclient.py
bffrost: I've changed my mind - I think my request here is unnecessary scope creep. I can add a separate, simple test for registering subsequent lots. There's enough going on in this test already.

bffrost:
Underneath the whitespace trimming issue there was another bug that actually was registering duplicate aliases. I fixed that here: https://github.com/mcneilco/acas-roo-server/pull/402 and now the tests are passing.

bffrost (approved):
LGTM now the test is passing!

dalejerikson:
Thanks for helping and taking care of this while I was on vacation :)

PR #104, opened 2022-11-29
ACAS-507: Add edit parent to acasclient

bffrost:
## Description
- Add `edit_parent` method which does both validation and actual edits. I discovered that the updateParent route doesn't perform validation, so I made this method always run validation before calling updateParent. Added a few tests confirming valid edits go through, and edits that would create duplicates are rejected.
- Tried to confirm that edit parent is an admin-only function and discovered it was not, so made an ACAS change (https://github.com/mcneilco/acas/pull/1047)
- Also updated swap_structures test to verify it now is admin-only
## How Has This Been Tested?
Ran ACASClient tests locally using ACAS branch `ACAS-507`
(see https://github.com/mcneilco/acas/pull/1047)

### DIFF
--- a/acasclient/acasclient.py
+++ b/acasclient/acasclient.py
@@ -2032,3 +2032,17 @@ class client():
         
         # Persist the update
         self.save_meta_lot(meta_lot)
+    
+    def edit_parent(self, parent, dry_run=True) -> Dict:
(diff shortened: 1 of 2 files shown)


brianbolt (approved):
LGTM

PR #105, opened 2022-11-30
ACAS-537: Enhance logging for unexpected experiment load failures

bffrost:
## Description
I noticed while debugging some upmerging issues that experiment loads were failing and giving very little clue to the cause of failure. In my case they were real, persistent issues that I was able to reproduce locally. However we see the same symptom in our occasional CI failures.

- Added an `expect_failures` flag to the `experiment_load_test` helper, so that expected failures are still treated the same, but unexpected failures will raise a clear exception containing the error messages sent back from ACAS.
- Updated all tests that expect failures

## How Has This Been Tested?
While troubleshooting, confirmed this yields a sensible and readable error message when unexpected failures happen.

After these changes, confirmed all ACASClient tests pass locally.

### DIFF
--- a/tests/test_acasclient.py
+++ b/tests/test_acasclient.py
@@ -417,7 +417,7 @@ class BaseAcasClientTest(unittest.TestCase):
         return user_client
 
     # Helper for testing an experiment upload was successful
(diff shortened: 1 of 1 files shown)


brianbolt (approved):
LGTM

PR #107, opened 2022-12-06
ACAS-550: Hopeful fix for flakey edit_parent test

bffrost:
## Description
- Could not reproduce the issue locally, but I'm making a guess that the problem is that the "get_meta_lot" call is happening so soon after the edit_parent call that it hasn't had time to commit the change to the parent. Testing out a fix which is to add a 50ms wait after the edit parent call to give time for the commit to happen before re-fetching the metalot.

## How Has This Been Tested?
Local tests still pass, but the real test will be on this PR.

### DIFF
--- a/tests/test_acasclient.py
+++ b/tests/test_acasclient.py
@@ -3,6 +3,7 @@
 """Tests for `acasclient` package."""
 
 from functools import wraps
+from time import sleep
 import unittest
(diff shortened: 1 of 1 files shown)


bffrost:
Closing since I opened this PR against the wrong branch

PR #108, opened 2022-12-06
ACAS-550: Fix flakey edit_parent test

bffrost:
## Description
- Could not reproduce the issue locally, but I'm making a guess that the problem is that the "get_meta_lot" call is happening so soon after the edit_parent call that it hasn't had time to commit the change to the parent. Testing out a fix which is to add a 50ms wait after the edit parent call to give time for the commit to happen before re-fetching the metalot.

## How Has This Been Tested?
Local tests still pass, but the real test will be on this PR.

### DIFF
--- a/tests/test_acasclient.py
+++ b/tests/test_acasclient.py
@@ -3,7 +3,6 @@
 """Tests for `acasclient` package."""
 
 from functools import wraps
-from time import sleep
 import unittest
 from acasclient import acasclient
 from pathlib import Path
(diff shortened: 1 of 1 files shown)


brianbolt (approved):
LGTM as a temporary workaround if it fixes the flakiness.  However, it seems weird that there is an async issue here. 

Is it possible that it's just not passing the validation step here:
https://github.com/mcneilco/acasclient/blob/b1d6b0ea01e44522687f85284c2c923dc1a164dd/acasclient/acasclient.py#L2054

Perhaps if it continues to fail we add some logic to check that the edit_parent step was successful.

brianbolt:
I see you are still working on it, I'll leave you to it.

bffrost:
@brianbolt  I didn't nail down the cause but you were definitely right that the edit was failing. I'm going to merge this to at least give us a better signal and better logging when this failure happens in the future.

PR #110, opened 2022-12-16
ACAS-567: add test for partially populated parent aliases

bffrost:
## Description
- This adds a test for a situation where empty-string aliases were getting created
- This goes alongside https://github.com/mcneilco/acas-roo-server/pull/410

## How Has This Been Tested?
- Ran acasclient tests locally on bbchem and indigo, with `client.cmpdreg.metaLot.allowDuplicateParentAliases=true` and `client.cmpdreg.metaLot.allowDuplicateParentAliases=false`


brianbolt (approved):
LGTM

PR #117, opened 2023-01-09
ACAS-577: Get all lots and filter by project code

brianbolt:
## Description
 - Simple change to get all lots which allows you to filter by a list of project codes

## Related Issue
ACAS-577

## How Has This Been Tested?
acasclient tests

PR #118, opened 2023-01-09
ACAS-578

brianbolt:
To be reviewed after #117 

## Description
 - Adds get_all_meta_lots to get all meta lots and optionally filter by project code
 - Adds get_meta_lots_by_lot_corp_names to get all metalots by corp name

## Related Issue
ACAS-578

## How Has This Been Tested?
acasclient tests

### DIFF
--- a/acasclient/acasclient.py
+++ b/acasclient/acasclient.py
@@ -1,5 +1,6 @@
 """Main module."""
 
+from concurrent.futures import ThreadPoolExecutor
 import requests
 import logging
 import os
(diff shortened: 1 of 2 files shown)


dalejerikson (approved):
👍

acasclient/acasclient.py
```diff
@@ -417,7 +421,51 @@ def get_all_lots(self):
...
+        Args:
+            lot_corp_names (list): A list of lot corp names
+
+        Returns: Returns an array of dict objects respresenting meta lots
+        """
+        meta_lots = []
+
+        # Call get_meta_lot for each lot corp name
+        # Run max_workers requests in parallel
+        with ThreadPoolExecutor(max_workers=max_workers) as executor:
```
bffrost, line 465: This is okay for now, but we may want to implement this on the server side in a single route if we start seeing performance issues.

bffrost (approved):

bffrost:
@brianbolt the automated CI tests are failing for this PR, and it does seem related to your new test. So please take a look and fix that before merging.

acasclient/acasclient.py
brianbolt: Yeah, I didn't want to mess with these routes or spend too much time on it.  I'll make a comment in the code about it.

brianbolt:
Fixed the tests with a workaround.  It worked locally because I was running the test independent of all other tests and the issue only came up when running in conjunction with other bulk load files tests.  I filed ACAS-581 to address the underlying issue that I workedaround.

PR #119, opened 2023-01-10
ACAS-579

brianbolt:
To be reviewed after #118 

## Description
 - Add method and tests for `get_lot_corp_names_by_bulk_load_file` which fetches lot corp names associated with a bulk load file
 - Add method and tests for `get_sdf_by_bulk_load_file` which fetches the SDF of a given bulk load id

## Related Issue
ACAS-579

## How Has This Been Tested?
acasclient tests

### DIFF
--- a/acasclient/acasclient.py
+++ b/acasclient/acasclient.py
@@ -1,5 +1,6 @@
 """Main module."""
 
+from concurrent.futures import ThreadPoolExecutor
 import requests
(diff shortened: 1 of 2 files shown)


brianbolt:
Tests are failing because there is a change in acas base to support these changes.

bffrost (approved):
LGTM

PR #122, opened 2023-01-11
ACAS-432: Import fileio/acas.py as selfile.py

brianbolt:
## Description
DataTeam Modules fileio/acas.py into acasclient.

I made minimal changes to the classes to get things working in acasclient package.  I left a PR review to highlight any notable changes.  I did not include the tests in this PR as there is some cleanup to do of proprietary info

## Related Issue
ACAS-432

## How Has This Been Tested?
Locally I ported the pytests to unittests and incorporated them into the acasclient test suite.

### DIFF
--- a/acasclient/selfile.py
+++ b/acasclient/selfile.py
@@ -4,8 +4,6 @@ Loader' files.
 Common formats are supported for both input and output: csv, and
 excel
 """
-# Contributors:  Shawn Watts
-
 
 ################################################################################
 # Packages
--- a/acasclient/selfile.py
+++ b/acasclient/selfile.py
@@ -433,7 +433,7 @@ class Generic(AbstractExperiment):
     and excel
     API examples:
     # Read an existing file, change the 'Project' metadata and save.
-    import datateam.fileio.acas as acas
+    import acasclient.selfile as acas
     expt = acas.Generic('35250_FG_acas.csv')
     expt.project = 'Test Project'
     expt.saveAs('35250_FG_modified.xlsx')
@@ -451,7 +451,7 @@ class Generic(AbstractExperiment):
     expt.page = 'page 19'
     expt.assay_date = '2020-05-20'
     expt.project = 'Star89'
-    # Dummy data, but this could come from a ctfile.sd_to_df, or
+    # Dummy data, but this could come from
     # rdkit.PandasTools, or whatever.
     expt.calculated_results_df = pd.DataFrame(
         {
--- a/acasclient/selfile.py
+++ b/acasclient/selfile.py
@@ -318,8 +318,6 @@ class AbstractExperiment():
     def experiment_name(self, value):
         self.meta_has_changed = (value != self._experiment_name)
         self._experiment_name = value
-    # Alias
-    name = experiment_name
 
 
     @property
--- a/acasclient/selfile.py
+++ b/acasclient/selfile.py
@@ -1,0 +1,851 @@
+"""
+A module to read, edit, and write ACAS 'Single Experiment
+Loader' files.
+Common formats are supported for both input and output: csv, and
+excel
+"""
+# Contributors:  Shawn Watts
+
+
+################################################################################
+# Packages
+################################################################################
+import collections
+import logging
+import io
+import os
+from functools import partial
+import pandas as pd
+
+
+################################################################################
+# Globals/Constants
+################################################################################
+# File formats/types.
+CSV = '.csv'
+XLS = '.xls'
+XLSX = '.xlsx'
+LOT_CORP_NAME = 'Corporate Batch ID'
+GENERIC = 'Generic'
+DOSE_RESPONSE = 'Dose Response'
+
+# ACAS-CmpdReg constants in sdf files.
+PARENT_CORP_NAME = 'Parent Corp Name'
+PARENT_STEREO_CAT = 'Parent Stereo Category'
+PARENT_ALIASES = 'Parent Aliases'
+LOT_NUM = 'Lot Number'
+LOT_NOTEBOOK_PAGE = 'Lot Notebook Page'
+LOT_CHEMIST = 'Lot Chemist'
+LOT_DATE = 'Lot Synthesis Date'
+LOT_COMMENTS = 'Lot Comments'
+LOT_SUPPLIER = 'Lot Supplier'
+LOT_SUPPLIER_ID = 'Lot Supplier ID'
+LOT_BARCODE = 'Lot Barcode'
+LOT_AMOUNT = 'Lot Amount'
+LOT_AMOUNT_UNITS = 'Lot Amount Units'
+LOT_PURITY = 'Lot Purity'
+LOT_PURITY_MEASURED_BY = 'Lot Purity Measured By'
+LOT_SALT_ABBREV = 'Lot Salt Abbrev'
+LOT_SALT_EQ = 'Lot Salt Equivalents'
+REG_LOT_CORP_NAME = 'Registered Lot Corp Name'
+REG_PARENT_CORP_NAME = 'Registered Parent Corp Name'
+REG_LOT_AMOUNT =  'Registered Lot Amount'
+REG_LOT_AMOUNT_UNITS =  'Registered Lot Amount Units'
+REG_PARENT_ALIASES = 'Registered Parent Aliases'
+REG_LEVEL = 'Registration Level'
+
+
+################################################################################
+# Logging
+################################################################################
+logger = logging.getLogger(__name__).addHandler(logging.NullHandler())
+
+
+################################################################################
+# Functions
+################################################################################
+def get_file_type(file_name):
+    """
+    :return: The file extension
+    :rtype: str
+    :param file_name: Path to file to check.
+    :type file_name: str
+    :note: a guess at the file format based on the extension:  csv, xls, or xlsx.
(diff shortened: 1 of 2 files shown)


acasclient/selfile.py
```diff
@@ -0,0 +1,851 @@
...
+# File formats/types.
+CSV = '.csv'
+XLS = '.xls'
+XLSX = '.xlsx'
+LOT_CORP_NAME = 'Corporate Batch ID'
+GENERIC = 'Generic'
+DOSE_RESPONSE = 'Dose Response'
+
+# ACAS-CmpdReg constants in sdf files.
+PARENT_CORP_NAME = 'Parent Corp Name'
```
brianbolt, line 33: I decided to leave these here for 2 reasons even though they aren't currently used within the acasclient package.

1. They could and should be used in the future but I didn't want to refactor other tests to use them right now.
2. By including them, the data team should be able to rely on pulling these from acasclient and remove their acas.py file

acasclient/selfile.py
```diff
@@ -0,0 +1,851 @@
...
+REG_LOT_AMOUNT =  'Registered Lot Amount'
+REG_LOT_AMOUNT_UNITS =  'Registered Lot Amount Units'
+REG_PARENT_ALIASES = 'Registered Parent Aliases'
+REG_LEVEL = 'Registration Level'
+
+
+################################################################################
+# Logging
+################################################################################
+logger = logging.getLogger(__name__).addHandler(logging.NullHandler())
```
brianbolt, line 61: Changed logger as per:https://docs.python.org/3/howto/logging.html#configuring-logging-for-a-library

acasclient/selfile.py
```diff
@@ -0,0 +1,851 @@
...
+    @property
+    def experiment_name(self):
+        """Experiment Name (Meta Data)"""
+        return self._experiment_name
+    @experiment_name.setter
+    def experiment_name(self, value):
+        self.meta_has_changed = (value != self._experiment_name)
+        self._experiment_name = value
+    # Alias
+    name = experiment_name
```
brianbolt, line 324: I added this alias for experiment name as we have a use case where all classes used in an ETL script are required to include a name property

acasclient/selfile.py
```diff
@@ -0,0 +1,851 @@
...
+    def project(self):
+        """Project (Meta Data)"""
+        return self._project
+    @project.setter
+    def project(self, value):
+        self.meta_has_changed = (value != self._project)
+        self._project = value
+
+    @property
+    def id(self):
```
brianbolt, line 377: I added id property as we have a use case where all classes used in an ETL script are required to include a id property

bffrost (approved):
LGTM, I gave this a light review with a focus on the areas that you've changed.

acasclient/selfile.py
shawn-schrodinger: @bffrost should 'name' a formal property (with a setter/getter)?

acasclient/selfile.py
shawn-schrodinger: 'id' is reserved name in python

acasclient/selfile.py
```diff
@@ -0,0 +1,851 @@
...
+        :type file_name: str
+        """
+
+        # Must provide either a file_name or a file_type but not both.
+        if file_name and file_type:
+            raise ValueError('Must provide either a file_name or a file_type but not both.')
+
+        # If file_name is provided, use it.
+        if file_name:
+            file_type = get_file_type(file_name)
```
shawn-schrodinger, line 544: Does this need and 'else' clause?

brianbolt: sorry forgot to note this change to saveAs.  I needed a method to get the file as text returned so I added the file_type as a parameter.  This shouldn't need an else because above we test if file_name and file_type are provided and throw an error.  Later file_name which is None causes the `df_out.to_csv` and `df_out.to_excel` to return their output instead of writing to a file.

acasclient/selfile.py
brianbolt: I guess I should test the setter but adding this as a setter did work.

acasclient/selfile.py
```diff
@@ -0,0 +1,851 @@
+"""
+A module to read, edit, and write ACAS 'Single Experiment
+Loader' files.
+Common formats are supported for both input and output: csv, and
+excel
+"""
+# Contributors:  Shawn Watts
```
shawn-schrodinger, line 7: this should be dropped.

acasclient/selfile.py
shawn-schrodinger: Alex and I actually discussed __slots__ vs decorators so the end user can not add more attributes.  I think adding more properties that don't map to a feature in an ACAS SEL file could lead to confusion.

bffrost: @brianbolt not to overcomplicate it, but you could implement the "name" field by wrapping this Experiment class in the ETL codebase. That'd keep acasclient "clean" since it is a little funny to have redundant names for the same attribute.

acasclient/selfile.py
bffrost: Really? https://realpython.com/lessons/reserved-keywords/
It is not a python keyword. What do you mean by it being a reserved name?

acasclient/selfile.py
brianbolt: All makes sense. I removed it here.

acasclient/selfile.py
shawn-schrodinger: Apologies, I meant to call attention that you are shadowing a *keyword*.  
https://www.programiz.com/python-programming/keyword-list

brianbolt: id is not listed on that page anywhere that I can find.

acasclient/selfile.py
```diff
@@ -0,0 +1,847 @@
...
+    # Create new, completely empty handle, and populate it.
+    expt = acas.Generic()
+    expt.protocol_name = 'my super cool protocol'
+    expt.experiment_name = 'ATP-Glo Assay 2020-05-19'
+    expt.scientist = 'Jackson'
+    expt.notebook = 'results_2020-05-19.pdf'
+    expt.page = 'page 19'
+    expt.assay_date = '2020-05-20'
+    expt.project = 'Star89'
+    # Dummy data, but this could come from a ctfile.sd_to_df, or
```
shawn-schrodinger, line 454: I think the API examples are good to keep, and verify.  But I realize that this ctfile module references something that is only our data team library.  Not critical, but removing the reference to ctfile might reduces questions from users down the line.  You may also want to consider lifting that code too, someday.  Since any other (non-text/clob) SD reader is almost certainly lossy in some, subtle, hard to debug, way.

acasclient/selfile.py
brianbolt: Consensus here seems to be that it's ok: 
https://stackoverflow.com/questions/3497883/in-python-is-it-bad-to-create-an-attribute-called-id

acasclient/selfile.py
```diff
@@ -0,0 +1,847 @@
...
+    represented as separate data members.  There is some internal
+    tracking to detect if the original data content is altered, but
+    the datatype changes need to be updated by the caller.
+    calculated_results_datatype - Datatype endpoint mapping.
+    calculated_results_df - Calculated Results dataframe.
+    Common formats are supported for both input and output:  csv,
+    and excel
+    API examples:
+    # Read an existing file, change the 'Project' metadata and save.
+    import datateam.fileio.acas as acas
```
bffrost, line 436: Shawn's right about the API examples referencing datateam libraries. We should also remove this line

acasclient/selfile.py
brianbolt: Thanks, make sense to look at lifting that code too in the future.  I modified this line slightly to remove the reference to ctfile.sd_to_df

acasclient/selfile.py
brianbolt: Thanks, I fixed it to import from acasclient.selfile

acasclient/selfile.py
brianbolt: Moving this PR to a draft as there are some more changes I want to make to the saveAs.  I realize my changes would break in the case of asking for an excel file_type as to_excel requires a file like object.

acasclient/selfile.py
```diff
@@ -0,0 +1,847 @@
...
+    :rtype: str
+    :param file_name: Path to file to check.
+    :type file_name: str
+    :note: a guess at the file format based on the extension:  csv, xls, or xlsx.
+    :raise: ValueError if the file extension is not recognized.
+    """
+
+    try:
+        ext = os.path.splitext(file_name)[1].lower()
+    except TypeError as err:
```
shawn-schrodinger, line 77: FYI - I just fixed a bug in our repo related to this.  Sometimes the filename stored in the ACAS experiment meta data is None, which raises an uncaught exception type.  My hot-fix is here: https://github.com/schrodinger/data_team_modules/pull/196
tested against all acas experiments in newt - and gracefully handled two cases that would have triggered this bug.

acasclient/selfile.py
shawn-schrodinger: Would df_out.to_csv() work, writing to a stringIo object as your file handle?

bffrost:
@brianbolt I'm closing this PR since it looks super stale

PR #126, opened 2023-01-27
ACAS-600 Add acasclient test for maxAutoLotNumber and reparent Lot

bffrost:
## Description
Add a test around the maxAutoLotNumber config. This test covers three issues:
1. Bulk loading a new lot of a compound that already has a high lot number will start numbering from the bottom
2. ACAS-600 NullPointerException when trying to reparent onto a parent that _only_ has a high-numbered lot
3. Reparenting onto a parent with both a high and a low lot number will respect autoLotNumber and increment up from the low lot number

## How Has This Been Tested?
- Tested locally without fix for ACAS-600 and confirmed I saw the exception
- Tested locally after ACAS-600 roo fix (https://github.com/mcneilco/acas-roo-server/pull/416) and test passed.


tests/test_acasclient.py
```diff
@@ -3674,6 +3687,80 @@ def test_010_validate_sdf_with_empty_aliases(self):
...
+                    "dbProperty": "Parent Stereo Category",
+                    "defaultVal": STEREO_CATEGORY,
+                    "required": True,
+                    "sdfProperty": None
+                }
+            ]
+        response = self.client.register_sdf(file_a, "bob", mappings, dry_run=False)
+        self.assertIn('New compounds: 1', response['summary'])
+        # Confirm it got lot 4000, and extract the parent number
+        registered = read_registered_csv(response)
```
brianbolt, line 3726: ah cool that's a nice one to have because there are lots of tests where I fetch all lots and sort them just to get the one I loaded.

brianbolt (approved):
LGTM

bffrost:
Confirmed my test passes after merging https://github.com/mcneilco/acas-roo-server/pull/416

PR #127, opened 2023-01-31
DISCO-42: Change to support Project Migration code

brianbolt:
## Description
 - Includes set file from data team
 - Experiment from another repository used for syncing ACAS data
 - Simple classes for AdditionalScientist
 - StereoCategory class
 - 

## Related Issue
DISCO-42

## How Has This Been Tested?
Ran project migration using changes in this PR


acasclient/CmpdReg.py
```diff
@@ -0,0 +1,196 @@
...
+        "lot_supplier_lot": meta_lot["lot"]["supplierLot"],
+        "lot_salt_abbrev": '; '.join(salt_abrevs) if len(salt_abrevs) > 0 else None,
+        "lot_salt_equivalents": '; '.join(salt_equivs) if len(salt_equivs) > 0 else None,
+        "lot_barcode": lot_barcode
+    }
+    return return_dict
+
+
+def convert_dict_to_type(dict, type_name):
+    """Converts a dictionary to a type."""
```
bffrost, line 112: Could use a little more context here on what this is used for, and what is meant by "type" here. Python type hints for the input arguments would be helpful.

acasclient/CmpdReg.py
```diff
@@ -0,0 +1,196 @@
...
+
+    def __init__(self, id=None, code=None, name=None, ignored=None):
+        super().__init__(AdditionalScientistType.ASSAY, id, code, name, ignored)
+
+
+def meta_lots_to_dict_array(meta_lots):
+    return [meta_lot_to_dict(meta_lot) for meta_lot in meta_lots]
+
+
+def meta_lot_to_dict(meta_lot):
```
bffrost, line 61: This could use a docstring, even though I get the idea of what this is doing

acasclient/CmpdReg.py
```diff
@@ -0,0 +1,196 @@
...
+        }
+    new_type = type(type_name, (object,), dict)()
+
+    # Add the as_dict method to the new type
+    new_type.as_dict = types.MethodType( as_dict, new_type )
+    return new_type
+
+
+def convert_object_to_sdf(object):
+    """Converts an object to an SDF."""
```
bffrost, line 134: This one also could use more detail. What type of object is expected here?

bffrost (commented):
Mostly LGTM. I didn't do a close review of `selfile.py` this time but I did check for anything proprietary which I didn't find.

Added some requests for more info about some functions that have either no docstring or very minimal docstrings. If you can add a bit more detail there, I'll give the approval

brianbolt:
@bffrost I added a bunch of typing and doc strings. Take a look. Some of this stuff is a little weird and experimental to make it fit the with project migration requirement that each data class be a python object with specific properties and methods like "as_dict"

bffrost (approved):
LGTM, thanks for the added comments and types

PR #128, opened 2023-02-10
Fixing some imports paths and naming

brianbolt:
## Description
 - The last PR for this issue had a few issues with paths and imports. Honestly not sure how I had it working in my local development environment but may have been some local install overriding what I thought I was using.
 - 

## Related Issue
DISCO-42

## How Has This Been Tested?
Ran import of acasclient using another project and verified it could be imported so the paths and class names are correct now.


bffrost (approved):
LGTM

PR #133, opened 2023-05-24
ACAS-672: Optional alias type and kind to add_parent_alias

brianbolt:
## Description
 - Adds optional type and kind to add_parent_alias

## Related Issue
ACAS-672

## How Has This Been Tested?
Ran acasclient tests
Used the method in some code and verified in the UI that the proper type and kind was set


bffrost (approved):
LGTM

PR #136, opened 2023-06-06
ACAS-679: Simple experiment loader class updates

brianbolt:
To be reviewed after ACAS-675

## Description
 - Updated simple experiment classes to pull in fixes which updated pandas and fix issue with raw results parsing.

## Related Issue
ACAS-679

## How Has This Been Tested?
Ran acasclient tests


acasclient/selfile.py
```diff
@@ -682,15 +759,17 @@ def raw_results(self, new_dict):
     def raw_results_datatype(self):
         """OrderedDict {(column1: temp_id1), ...)} column name keys for raw results data."""
         return self._raw_results
+
     @raw_results.setter
-    def raw_results(self, new_dict):
+    def raw_results_datatype(self, new_dict):
```
bffrost, line 764: I'm confused by this function and its new name.
Do we have two `@raw_results.setter` functions? One on this line and another on line 754 above?
It seems like prior to your change we just had duplicated code. After your change we have a function named `raw_results_datatype` that is annotated as a setter for the `raw_results` property, not the `raw_results_datatype` property... and the code for both updates the `self._raw_results` dictionary.

bffrost (changes requested):
Aside from one point of confusion around `raw_results_datatype` this looks good. Mostly just reformatting, plus replacing `append` with `concat` and adding Excel handling when no file name is set.

Please address my confusion around `raw_results_datatype` then I'll give the approval.

acasclient/selfile.py
bffrost: This looks better.

bffrost (approved):
LGTM!

PR #137, opened 2023-07-20
ACAS-674: Order corporate name based on similarity match.

iamanandkr:
See https://github.com/mcneilco/acas-roo-server/pull/445 for details. I am here adding unit tests for the workflow.

## How Has This Been Tested?

TDD.


tests/test_acasclient.py
```diff
@@ -1228,10 +1234,17 @@ def test_008_cmpd_search(self):
         search_results = self.client.\
             cmpd_search(molStructure=molStructure)
         self.assertGreater(len(search_results["foundCompounds"]), 0)
+
         # Search by ID list
-        corp_name_list = ["CMPD-0000001"]
+        lots = self.client.get_all_lots()
+        corp_name_list = [lot['lotCorpName'] for lot in lots]
+        corp_name_list = ["CMPD-005"]
```
brianbolt, line 1241: I'm not positive, but I can imagine this test becoming flaky.  You loop through all lots and create a list of lot car names and then replace that list with a hardcoded string.  I'd suggest fetching one of the lots parent for names from the list of lots and replacing a few of the "0000" in the string to get your corp name list.  Then below when you assert that you get the one compound, you can re-use the original parent identifier.

brianbolt (changes requested):

tests/test_acasclient.py
iamanandkr: Done. I initially considered that but as an alternative I ended up using `@requires_absent_basic_cmpd_reg_load` decorator to delete all the existing experiment data to ensure `CMPD-0000005` would be added from the new test file.

PR #142, opened 2023-10-21
ACAS-711: Tests for downloading cmpdreg file content

brianbolt:
## Description
 - Tests for fixing issue with downloading cmpdreg files with uppercase (or mixed case) extensions
 - DRYed up some file content comparisons

## Related Issue
ACAS-711

## How Has This Been Tested?
ran acasclient tests before and after applying acas-roo-server fixes

### DIFF
--- a/acasclient/acasclient.py
+++ b/acasclient/acasclient.py
@@ -440,23 +440,23 @@ class client():
         """
         return self.add_files_to_lot(lot_corp_name, [{"file": file, "file_type": file_type, "writeup": writeup}])
 
-    def _get_cmpdreg_files_content(self, files):
+    def _get_cmpdreg_files_content(self, cmpdreg_lot_file_list):
         """Get the content of cmpdreg files
 
-        Pass an array of cmpdreg_files to download and fill their content attribute
+        Pass the lot fileList array of array attribute of a cmpdreg metalot object to update or add the content attribute to each file
 
         Args:
-            files: An array of cmpdreg_files to download in the format (attributes other than url are ignored)
+            files: An array of dictionaries from the lot fileList attribute (attributes other than url are ignored)
                 "url": The url of the file to download
 
         Returns:
-            The same array of cmpdreg_files with the "content" attribute added
+            The the original array of dictionaries with the content attribute updated or added to each file
         """
-        for file in files:
+        for file in cmpdreg_lot_file_list:
             resp = self.session.get("{}/cmpdreg/MultipleFilePicker/{}".format(self.url, file["url"]))
             resp.raise_for_status()
             file["content"] = resp.content
-        return files
+        return cmpdreg_lot_file_list
 
     def _upload_cmpdreg_files(self, lot_corp_name, files):
         """Upload a list of files to CmpdReg.
--- a/acasclient/acasclient.py
+++ b/acasclient/acasclient.py
@@ -440,6 +440,25 @@ class client():
         """
         return self.add_files_to_lot(lot_corp_name, [{"file": file, "file_type": file_type, "writeup": writeup}])
 
+    def _get_cmpdreg_files_content(self, files):
+        """Get the content of cmpdreg files
+
+        Pass an array of cmpdreg_files to download and fill their content attribute
+
+        Args:
+            files: An array of cmpdreg_files to download in the format
+                "subdir": The subdir of the file (usually lot_corp_name)
+                "url": The url of the file to download
+
+        Returns:
+            The same array of cmpdreg_files with the "content" attribute added
+        """
+        for file in files:
+            resp = self.session.get("{}/cmpdreg/MultipleFilePicker/{}".format(self.url, file["url"]))
+            resp.raise_for_status()
+            file["content"] = resp.content
+        return files
+
     def _upload_cmpdreg_files(self, lot_corp_name, files):
         """Upload a list of files to CmpdReg.
 
--- a/acasclient/acasclient.py
+++ b/acasclient/acasclient.py
@@ -446,8 +446,7 @@ class client():
         Pass an array of cmpdreg_files to download and fill their content attribute
 
         Args:
-            files: An array of cmpdreg_files to download in the format
-                "subdir": The subdir of the file (usually lot_corp_name)
+            files: An array of cmpdreg_files to download in the format (attributes other than url are ignored)
                 "url": The url of the file to download
 
         Returns:
--- a/tests/test_acasclient.py
+++ b/tests/test_acasclient.py
@@ -437,6 +437,16 @@ class BaseAcasClientTest(unittest.TestCase):
         user_client = acasclient.client(user_creds)
         return user_client
 
+    def verify_file_and_content_equal(self, content, file):
+        """ Compare the content to the file contents"""
+        if isinstance(content, bytes):
+            with open(file, 'rb') as f:
+                upload_bytes = f.read()
+                self.assertEqual(content, upload_bytes)
+        else:
+            with open(file, 'r') as f:
+                self.assertEqual(content, f.read())
+
     # Helper for testing an experiment upload was successful
     def experiment_load_test(self, data_file_to_upload, dry_run_mode, expect_failure=False, report_file_to_upload=None, images_file_to_upload=None):
         response = self.client.\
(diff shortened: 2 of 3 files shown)


acasclient/acasclient.py
```diff
@@ -440,6 +440,25 @@ def add_file_to_lot(self, lot_corp_name, file, file_type, writeup=None):
         """
         return self.add_files_to_lot(lot_corp_name, [{"file": file, "file_type": file_type, "writeup": writeup}])
 
+    def _get_cmpdreg_files_content(self, files):
```
brianbolt, line 443: file['url'] shows the exact location on the server of the server and the route in acas to fetch that file requires the same full path.  So instead of simple parameters like "lot_corp_name", "file_name" we have to pass the actual url.  Because of this, I just decided to make this an "_" function to fill in the content of the file list returned in metalots saves.

tests/test_acasclient.py
```diff
@@ -437,6 +437,16 @@ def create_and_connect_backdoor_user(self, username = None, password = None, pre
         user_client = acasclient.client(user_creds)
         return user_client
 
+    def verify_file_and_content_equal(self, content, file):
```
brianbolt, line 440: Made a function to DRY up some of this testing.

tests/test_acasclient.py
```diff
@@ -466,27 +476,16 @@ def experiment_load_test(self, data_file_to_upload, dry_run_mode, expect_failure
...
             # If the file['content'] is string then read the data_file_to_upload as string
             source_file = experiment.get_source_file()
-            if isinstance(source_file['content'], bytes):
-                with open(data_file_to_upload, 'rb') as f:
-                    upload_bytes = f.read()
-                    self.assertEqual(source_file['content'], upload_bytes)
-            else:
-                with open(data_file_to_upload, 'r') as f:
-                    self.assertEqual(source_file['content'], f.read())
+            self.verify_file_and_content_equal(source_file['content'], data_file_to_upload)
```
brianbolt, line 479: DRY usage

tests/test_acasclient.py
```diff
@@ -466,27 +476,16 @@ def experiment_load_test(self, data_file_to_upload, dry_run_mode, expect_failure
...
                 
-                if isinstance(report_file['content'], bytes):
-                    with open(report_file_to_upload, 'rb') as f:
-                        upload_bytes = f.read()
-                        self.assertEqual(report_file['content'], upload_bytes)
-                else:
-                    with open(report_file_to_upload, 'r') as f:
-                        self.assertEqual(report_file['content'], f.read())
+                # Verify matching file content
+                self.verify_file_and_content_equal(report_file['content'], report_file_to_upload)
```
brianbolt, line 488: DRY usage

tests/test_acasclient.py
```diff
@@ -3999,7 +3998,7 @@ def test_011_upload_cmpdreg_files(self):
         file_name1 = 'dummy.pdf'
         file_test_path1 = Path(__file__).resolve().parent\
             .joinpath('test_acasclient', file_name1)
-        file_name2 = 'dummy2.pdf'
+        file_name2 = 'dummy2.PDF' # Upper case on purpose to verify we can handle uppercase extensions
```
brianbolt, line 4001: This is the actual change to test, which is uploading a upper case extension

tests/test_acasclient.py
```diff
@@ -4102,14 +4101,20 @@ def test_012_upload_lot_files(self):
         self.assertIn("metalot", meta_lot_save_response)
         self.assertIn("fileList", meta_lot_save_response["metalot"])
 
+        # Fill the content of the files
+        saved_files = self.client._get_cmpdreg_files_content(meta_lot_save_response["metalot"]["fileList"])
```
brianbolt, line 4105: Use the new acasclient method to fill the 'content' with the file content.

tests/test_acasclient.py
```diff
@@ -4102,14 +4101,20 @@ def test_012_upload_lot_files(self):
...
+
         # Verify that the fileList has the files we just uploaded by checking the names
         has_file1 = False
         has_file2 = False
-        for file in meta_lot_save_response["metalot"]["fileList"]:
+        for file in saved_files:
             if file["name"] == file_name1:
                 has_file1 = True
+                self.verify_file_and_content_equal(file['content'], file_test_path1)
```
brianbolt, line 4113: Verify we match contents of the file we uploaded with what we downloaded.

tests/test_acasclient.py
iamanandkr, line 440: Can we replace `file` with `fpath` / `file_path` or something on those lines to indicate that this is supposed to be a file path and not a file handle :)

iamanandkr, line 440: `:nit` If the function name is `_file_and_content_` then I would expect the first argument to be the file path and the second argument as the content.

tests/test_acasclient.py
```diff
@@ -437,6 +437,16 @@ def create_and_connect_backdoor_user(self, username = None, password = None, pre
...
 
+    def verify_file_and_content_equal(self, content, file):
+        """ Compare the content to the file contents"""
+        if isinstance(content, bytes):
+            with open(file, 'rb') as f:
+                upload_bytes = f.read()
+                self.assertEqual(content, upload_bytes)
+        else:
+            with open(file, 'r') as f:
+                self.assertEqual(content, f.read())
```
iamanandkr, line 448: `nit:` You can also consider writing it as,
```python
mode = 'rb' if isinstance(content, bytes) else 'r'
with open(file, mode) as f:
    sefl.assertEqual(content, f.read())
```

acasclient/acasclient.py
```diff
@@ -440,6 +440,25 @@ def add_file_to_lot(self, lot_corp_name, file, file_type, writeup=None):
...
+        """Get the content of cmpdreg files
+
+        Pass an array of cmpdreg_files to download and fill their content attribute
+
+        Args:
+            files: An array of cmpdreg_files to download in the format
+                "subdir": The subdir of the file (usually lot_corp_name)
```
iamanandkr, line 450: Can we mention that this is optional? Or what do you think about totally removing it -  In the request we only pass the URL of the file, and we are returning the entire `files` object as it is with the "content" parameter added.

acasclient/acasclient.py
```diff
@@ -440,6 +440,25 @@ def add_file_to_lot(self, lot_corp_name, file, file_type, writeup=None):
...
+                "url": The url of the file to download
+
+        Returns:
+            The same array of cmpdreg_files with the "content" attribute added
+        """
+        for file in files:
+            resp = self.session.get("{}/cmpdreg/MultipleFilePicker/{}".format(self.url, file["url"]))
+            resp.raise_for_status()
+            file["content"] = resp.content
```
iamanandkr, line 459: Sanity check: If the user already has `content` as a key in the dictionary then we want to override it? 

If we don't want to do that, then doing validation check and raising an exception might be a good idea.

iamanandkr: Also instead of working on the actual input object, do we want to create a copy of it, update that and return that? I am just considering if users would find it weird that their input object got updated.

acasclient/acasclient.py
```diff
@@ -440,6 +440,25 @@ def add_file_to_lot(self, lot_corp_name, file, file_type, writeup=None):
         """
         return self.add_files_to_lot(lot_corp_name, [{"file": file, "file_type": file_type, "writeup": writeup}])
 
+    def _get_cmpdreg_files_content(self, files):
+        """Get the content of cmpdreg files
+
+        Pass an array of cmpdreg_files to download and fill their content attribute
+
+        Args:
+            files: An array of cmpdreg_files to download in the format
```
iamanandkr, line 449: I am not sure what a good name would be, but `files` doesn't seem to totally indicate what kind of  a data structure it is. I would think this would just a take list of file paths, but it's more than that - it's a list of dictionaries which contain the "url" key(maybe use type annotation to force users to read the documentation?).

Alternatively we could just keep this method to take an argument called: `urls` and return a list of dictionaries with `url` and `content` keys.

iamanandkr (commented):
Looks good.

API created in the client is a private API and it's being used for testing the feature, but I still made a few comments - please feel free to ignore those if this is a time crunch thing.

Should we also rename the `dummy2.pdf` file in the tests to be `dummy2.PDF`?

tests/test_acasclient.py
brianbolt: Thanks, thats a nice simplification

acasclient/acasclient.py
brianbolt: I think this is the one suggestion I won't take. I think if you pass into this function, you would expect it to override your content. I imagine a scenario where you may want to refresh the files objects from whats on server.  I also think its ok to update the object directly.  Not that memory is a problem here but I see this function as more of a utility function to fill out the content immediately after fetching files.  I could have also added this function to be called as an optional `fetch_file_content` parameter when fetching a metalot but I chose to keep that out for now.

iamanandkr (approved):
LGTM!

acasclient/acasclient.py
brianbolt: Thanks, point taken, it was confusing. Considering this is supposed to be the lot.fileList attribute, I just described it as that and anyone using it can see the structure that is supposed to be passed in now.

PR #144, opened 2023-10-24
Add test for vendorID updates

brianbolt:
## Description
 - Adds test for meta lot vendorID updates. See changes on https://github.com/mcneilco/acas-roo-server/pull/453

## Related Issue
https://jira.schrodinger.com/browse/ACAS-696

## How Has This Been Tested?
Ran acasclient tests and verified manually that vendorID updates persist

### DIFF
--- a/tests/test_acasclient.py
+++ b/tests/test_acasclient.py
@@ -3216,12 +3216,15 @@ class TestCmpdReg(BaseAcasClientTest):
 
         # Verify our cmpdreg admin user can save the restricted lot
         meta_lot["lot"]["color"] = "red"
+        VENDOR_ID = "12345"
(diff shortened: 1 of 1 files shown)


bffrost (approved):
LGTM

bffrost:
@brianbolt I noticed some test failures around "dummy2.PDF". I don't see how they could be related to this change, but could you take a look and see if you have theories about where that's coming from?

brianbolt:
> @brianbolt I noticed some test failures around "dummy2.PDF". I don't see how they could be related to this change, but could you take a look and see if you have theories about where that's coming from?

Yes, going to go ahead and merge this change, but I will investigate those errors.

PR #149, opened 2023-11-18
Attempted fixes for flaky tests

bffrost:
## Description
Observed two phenotypes of test failures:
- https://github.com/mcneilco/acas/actions/runs/6909971925/job/18802300896 where the SDF load for SEL fails with a mismatched corp names error, as if a compound has been left in the system when we try to re-load the basic file
- https://github.com/mcneilco/acas/actions/runs/6910041295/job/18802488125 where the edit parent test fails trying to fetch one of the metalots

I'd like to test out a fix for the first issue which is a more aggressive deletion of lots when the expected lots aren't found.
I'm really not sure the cause of the edit parent failure, but I want to investigate it further so I've uncommented the rest of the edit parent test to try to catch it failing when the tests run.

## How Has This Been Tested?
- Ran acasclient tests locally, but the real test will come from running & re-running in GHA

### DIFF
--- a/tests/test_acasclient.py
+++ b/tests/test_acasclient.py
@@ -288,6 +288,12 @@ def delete_all_lots_and_experiments(self):
     if len(lots) > 0:
         self.delete_all_experiments()
         self.delete_all_cmpd_reg_bulk_load_files()
+        # Check lots again
+        lots = self.client.get_all_lots()
+        if len(lots) > 0:
+            # Delete any remaining lots not from bulk loads
+            for lot in lots:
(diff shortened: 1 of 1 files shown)


brianbolt (commented):
I know its merged already but LGTM.

bffrost:
Sorry @brianbolt I was just seeing rampant failures so I merged this to try to squash them. I may have to do several rounds but it was failing so much I couldn't get the PR tests on this repo to run because the release-2024.1.x tag hasn't built successfully yet.

PR #151, opened 2023-12-06
ACAS-704 test project restrictions

brianbolt:
## Description
 - Adding tests for experiment project restrictions

## Related Issue
ACAS-712

## How Has This Been Tested?
Ran acasclient tests

### DIFF
--- a/tests/test_acasclient.py
+++ b/tests/test_acasclient.py
@@ -3343,7 +3343,7 @@ class TestCmpdReg(BaseAcasClientTest):
         file_to_upload = get_basic_experiment_load_file(self.tempdir, project.names[PROJECT_NAME], restricted_lot_corp_name, file_name='4 parameter D-R.csv')
         response = self.client.\
             experiment_loader(file_to_upload, "bob", False)
-        self.assertTrue(self.check_lot_exists_in_experiment(self, restricted_lot_corp_name, response['results']['experimentCode']))
+        self.assertTrue(self.check_lot_exists_in_experiment(restricted_lot_corp_name, response['results']['experimentCode']))
                         
         # Deny rule: Creg admin is not an acas user
         self.assertFalse(can_delete_lot(self, cmpdreg_admin, restricted_lot_corp_name, set_owner_first=True))
@@ -3359,7 +3359,7 @@ class TestCmpdReg(BaseAcasClientTest):
         self.assertIsNone(meta_lot)
 
         # Get the experiment and verify the lot is deleted
-        self.assertFalse(self.check_lot_exists_in_experiment(self, restricted_lot_corp_name, response['results']['experimentCode']))
+        self.assertFalse(self.check_lot_exists_in_experiment(restricted_lot_corp_name, response['results']['experimentCode']))
 
         # TODO: The following tests are dependent on a non default configuration: client.cmpdreg.metaLot.disableDeleteMyLots = false so we can't test this by default. We should turn this back on when we have a testing system which can change the configurations of acas.
         # # Global compound, with experiment in Global project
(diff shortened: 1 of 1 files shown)


dalejerikson (approved):
LGTM!

dalejerikson:
I see that in GitHub Actions that there's one error but I'm not sure if there's still problems with ACAS builds in Github or not...

`File "/home/runner/work/acasclient/acasclient/acasclient/tests/test_acasclient.py", line 3346, in test_004_delete_lot
    self.assertTrue(self.check_lot_exists_in_experiment(self, restricted_lot_corp_name, response['results']['experimentCode']))
TypeError: check_lot_exists_in_experiment() takes 3 positional arguments but 4 were given`

tests/test_acasclient.py
```diff
@@ -3343,7 +3343,7 @@ def check_lot_exists_in_experiment(self, lot_corp_name, experiment_code_name):
         file_to_upload = get_basic_experiment_load_file(self.tempdir, project.names[PROJECT_NAME], restricted_lot_corp_name, file_name='4 parameter D-R.csv')
         response = self.client.\
             experiment_loader(file_to_upload, "bob", False)
-        self.assertTrue(check_lot_exists_in_experiment(self, restricted_lot_corp_name, response['results']['experimentCode']))
+        self.assertTrue(self.check_lot_exists_in_experiment(self, restricted_lot_corp_name, response['results']['experimentCode']))
```
iamanandkr, line 3346: `self` shouldn't be passed now. It's the same for line 3362.

iamanandkr (dismissed):
Thank you Brian! 
I only have comment about the failing test, other than that LGTM!

tests/test_acasclient.py
```diff
@@ -4600,4 +4600,51 @@ def test_018_image_file(self):
...
+        # (Allowed) Restricted experiment proj1 with lot unrestricted
+        # Basic cmpdreg load creates a lot in the Global unrestricted project
+        # So we expect that it can be uploaded to a restricted project
+        # Get a file to upload to restricted_project_1 but has a lot from the Global unrestricted project
+        file_to_upload = get_basic_experiment_load_file(self.tempdir, restricted_project_1.names[PROJECT_NAME], corp_name=None, file_name='uniform-commas-with-quoted-text.csv')
+        # Upload the file and expect it to succeed
+        response = self.experiment_load_test(file_to_upload, True, expect_failure=False)
+
+        # (Allowed) Unrestricted experiment proj1 with lot restricted proj2
+        # Create an upload file which uploads to restricted_project_1 but has a lot from restricted_project_2
```
bffrost, line 4647: From the comment this looks like a good test to run. However I'm confused about your implementation. On the line below this, you seem to be loading a lot in restricted project 1 into experiment restricted project 1.
I don't see you create an unrestricted experiment project or using Global as the experiment project.
Also the lot seems to be from project 1 but your comment here says it's from project 2.

bffrost (changes requested):
Great tests to have. Last part of test 19 doesn't seem to line up with the code comment. Also see Anand's comment to fix the breaking test

tests/test_acasclient.py
iamanandkr: I have addressed the comment. Let me know if now the test seems correct to you!

bffrost (approved):
LGTM

tests/test_acasclient.py
brianbolt: Ah! Thanks.

tests/test_acasclient.py
brianbolt: Thanks. LGTM.  I changed the test and forgot to fix the comment.

PR #156, opened 2024-03-04
Add handling for new chemist attribute in parentLot/getAllAuthorizedLots route 

hhan-schrodinger:
## Description
Users requested including lot chemist as an attribute for each record in creg metadata when calling the cmpdReg/parentLot/getAllAuthorizedLots endpoint with the goal of helping analyze and categorize records by using lot chemist.

After updating backend, added docstring and test to check for the new attribute in related acasclient functions.


bffrost (approved):
LGTM

PR #163, opened 2024-06-04
ACAS-773: acasclient get_protocols_by_label failing when protocol has "/" in the name

brianbolt:
## Description
 - @bffrost Added tests for reproducing the issue including more complex special characters
 - @brianbolt Updated acasclient.py to use urllib.parse.quote for URL encoding to fix the tests

## Related Issue
ACAS-773

## How Has This Been Tested?
Ran acasclient tests


bffrost (approved):
LGTM

PR #165, opened 2024-07-29
Add ls_kinds property

brianbolt:
## Description
 - Adds ls_kinds property to experiment object which fetches the list of ls_kind from the experiment data columns experiment value.
 
## Related Issue
DISCO_1468

## How Has This Been Tested?
 - Ran the code using an ETL script which uses the information to detect when an experiment ls_kind exists.

### DIFF
--- a/acasclient/experiment.py
+++ b/acasclient/experiment.py
@@ -87,6 +87,15 @@ class Experiment(dict):
         return protocol_name["labelText"]
 
     @property
+    def ls_kinds(self) -> list:
+        data = get_entity_values_by_state_type_kind_value_type(
+                entity=self,
+                state_type="metadata",
+                state_kind="data column order",
+                value_type="codeValue")
+        return [d['codeValue'] for d in data if d['codeKind'] == 'column name']
+
+    @property
     def project(self) -> None:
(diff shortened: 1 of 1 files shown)


acasclient/experiment.py
```diff
@@ -86,6 +86,15 @@ def protocol_name(self) -> str:
                 label_kind="protocol name")
         return protocol_name["labelText"]
 
+    @property
+    def ls_kinds(self) -> list:
```
bffrost, line 90: @brianbolt My one issue here is that "ls_kind" is really an internal concept to ACAS, and won't be immediately understood by users of acasclient. I think calling the property `column_names` or `result_types` would be a lot clearer as to what these strings represent.

bffrost (changes requested):

acasclient/experiment.py
brianbolt: Thanks makes sense but how about `result_kinds`?  Only because result types is really numeric, string...etc. and column names would be the combination of "result_kind" + units

bffrost: I'm good with `result_kinds`

bffrost (approved):
LGTM

PR #166, opened 2024-09-18
DISCO-1674: column_information, result_types, is_images_file_experiment, image zip bug fixes

brianbolt:
## Description
 - This adds new functionality to the experiment object
   - `column_information` property to get all column information for an experiment
   - `result_types` property to get just the result types from the column information
   - `is_images_file_experiment` to check if inline file value result types were uploaded to the experiment
 - Speeds up creating an image file zip by first checking if the experiment is an image file experiment.
 - Fixes a bug where the client passed into the get_images_file function is not used.

## Related Issue
[DISCO-1674](https://schrodinger.atlassian.net/browse/DISCO-1674)

## How Has This Been Tested?
Added and ran acasclient tests

### DIFF
--- a/acasclient/experiment.py
+++ b/acasclient/experiment.py
@@ -123,11 +123,6 @@ class Experiment(dict):
 
     @property
     def result_kinds(self) -> list:
-        data = get_entity_values_by_state_type_kind_value_type(
-                entity=self,
-                state_type="metadata",
-                state_kind="data column order",
-                value_type="codeValue")
         return [d['codeValue'] for d in self.column_information if d['codeKind'] == 'column name']
 
     @property
--- a/acasclient/experiment.py
+++ b/acasclient/experiment.py
@@ -58,7 +58,7 @@ def fetch_analysis_groups_if_missing(func):
         return func(self, *args, **kwargs)
     return wrapper
 
-def fetch_analysis_groups_if_images_expertiment(func):
+def fetch_analysis_groups_if_images_experiment(func):
     """Decorator to fetch analysis groups if they are missing from the experiment dict
     """
     @wraps(func)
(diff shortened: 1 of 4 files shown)


acasclient/experiment.py
```diff
@@ -36,10 +37,35 @@ def fetch_analysis_groups_if_missing(func):
...
         if ANALYSIS_GROUPS_KEY not in self:
-            experiment_dict = self.client.get_experiment_by_code(self.code, full=True)
-            self.__init__(experiment_dict, client=self.client)
-        return func(self)
+            experiment_dict = aclient.get_experiment_by_code(self.code, full=True)
+            self[ANALYSIS_GROUPS_KEY] = experiment_dict[ANALYSIS_GROUPS_KEY]
+        return func(self, *args, **kwargs)
+    return wrapper
+
+def fetch_analysis_groups_if_images_expertiment(func):
```
bffrost, line 61: typo in `expertiment`, should be `experiment`

acasclient/experiment.py
```diff
@@ -86,14 +112,27 @@ def protocol_name(self) -> str:
...
+        data = get_entity_values_by_state_type_kind_value_type(
+                entity=self,
+                state_type="metadata",
+                state_kind="data column order",
+                value_type="codeValue")
+        return data
+
     @property
     def result_kinds(self) -> list:
         data = get_entity_values_by_state_type_kind_value_type(
```
bffrost, line 126: Can we remove this call now that you're getting the data from `self.column_information` rather than `data`?

bffrost (changes requested):
Overall this looks good but I've noticed a few bits to clean up and correct.

acasclient/experiment.py
brianbolt: Thanks fixed.

acasclient/experiment.py
brianbolt: Yes, thanks, missed removing that. Fixed.

PR #167, opened 2024-10-21
DISCO-1604: Protocol support, salt class, metalot structure fix and logging

brianbolt:
## Description
 - Protocol support
   - In roo protocol is a subclass of abstract thing so did the same thing when creating `Protocol` object
   - Overrode save route to create and update protocol ls things
   - Added Author code type to handle adding scientist name in Protocol object.
 - Salt class
   - Basic salt class which can be initiated and saved.
 - Changed meta lot to bulk load entry to prefer parent mol by default 
 - Fixed logging issue which was causing duplicate logging in python script which created their own basic logging configuration.

## Related Issue
DISCO-1604

## How Has This Been Tested?
Added acasclient tests for protocol changes
Verified duplicate login issue was fixed when using an external library.


acasclient/CmpdReg.py
```diff
@@ -80,8 +126,13 @@ def meta_lot_to_dict(meta_lot: dict) -> dict:
     # If lot amount is filled in, then we need to fill in lot barcode with the lot corp name
     lot_barcode = meta_lot["lot"]["corpName"] if meta_lot["lot"]["amount"] is not None else None
 
+    if use_parent_mol:
+        mol = meta_lot["lot"]["parent"]["molStructure"]
```
brianbolt, line 130: Changed the default to use the parent mol structure because this is used for ETL purposes where we do not want lots to split off based on the way they were drawn if the destination server has different ACAS chemistry versions.

bffrost: That makes sense. It's hard to say, but using the parent structure seems like reasonable default behavior as most folks aren't even aware of the asDrawnStruct at the lot level

acasclient/lsthing.py
```diff
@@ -850,7 +849,18 @@ def as_camel_dict(self):
...
         camel_dict = convert_json(snake_case_dict, underscore_to_camel)
         return camel_dict
+    
+    def __deepcopy__(self, memo):
+        # Create a deep copy of the instance first
+        copied_obj = copy.copy(self)  # Shallow copy of self
+        # Then deep copy the attributes individually
+        copied_obj.__dict__ = copy.deepcopy(self.__dict__, memo)
+        # Set the 'id' attribute to None
+        copied_obj.id = None
```
bffrost, line 859: While it makes sense to put this nulling out into this function, someone coming across this function wouldn't really have reason to expect it will clear out these specific keys. I think this function should have a docstring explaining that it has this special behavior.

acasclient/acasclient.py
```diff
@@ -960,7 +960,90 @@ def get_protocols_by_label(self, label):
...
+        """Get all protocols
 
+        Get an array of all protocols
+
+        Args:
+            None
+
+        Returns: Returns an array of protocols
+        """
+        resp = self.session.get("{}/api/protocolCodes"
```
bffrost, line 991: Does this API really return protocols, not protocol codes?
If I'm reading the code of `exports.protocolCodeList` correctly, it returns little codeTable objects of code, name, and ignored.

I find this acasclient endpoint to be confusingly named - I'd expect it to return full protocol objects, with the same structure that `get_protocol_by_code` returns.

bffrost (changes requested):
This looks reasonable, but the `get_all_protocols` endpoint naming compared to the contents of the response is confusing to me. Awaiting your response before approving

acasclient/acasclient.py
brianbolt: Thanks for pointing that out.  This was AI copy pasta that I did not catch. I just added the real return info in the doc string.

acasclient/lsthing.py
brianbolt: Thanks, yeah, just added a clearer doc string on this.

acasclient/acasclient.py
bffrost: That's helpful. I still do find the function name `get_all_protocols` misleading though. Maybe `get_all_protocol_codetables` or `get_all_protocol_stubs`?
Those dicts are not `Protocol`s are they?

acasclient/protocol.py
```diff
@@ -0,0 +1,70 @@
...
+PROTOCOL_STATUS = 'protocol status'
+ASSAY_STAGE = 'assay stage'
+SCIENTIST = 'scientist'
+ASSAY_PRINCIPLE = 'assay principle'
+CREATION_DATE = 'creation date'
+PROTOCOL_DETAILS = 'protocol details'
+NOTEBOOK = 'notebook'
+
+
+class Protocol(SimpleLsThing):
```
bffrost, line 25: It seems weird / wrong to me to have Protocol inherit from SimpleLsThing. I can see that we want to use the same pythonic dictionary representation of an AbstractThing object, but there is LsThing-specific stuff such as `first_ls_things` / `second_ls_things` in the SimpleLsThing class.

I know it's more work, but I'd prefer to see a refactor that creates an AbstractSimpleThing class to have the common logic between this Protocol class and true LsThing-table-based SimpleLsThings

acasclient/protocol.py
```diff
@@ -0,0 +1,70 @@
+from acasclient.lsthing import (CodeValue, SimpleLsThing, LsThing, clob, ACAS_DDICT, ACAS_AUTHOR)
+
+from datetime import datetime
+
+# Constants
+ACAS_LSTHING = 'ACAS LsThing'
```
bffrost, line 6: This constant doesn't seem to be referenced and seems misplaced here

acasclient/acasclient.py
brianbolt: Just updated it to be get_all_protocol_stubs

acasclient/protocol.py
```diff
@@ -0,0 +1,70 @@
...
+        :type client: acasclient.client
+        """
+        # Run validation
+        if not skip_validation:
+            self.validate(client)
+        self._prepare_for_save(client)
+        # Persist
+        protocol_dict = self._ls_thing.as_camel_dict()
+        resp_dict = client.save_protocol(protocol_dict)
+        self._ls_thing = LsThing.from_camel_dict(resp_dict)
```
bffrost, line 68: I can see this pattern exists elsewhere in the code, but it is strange to me to use the `LsThing` class here, when a Protocol is not an LsThing in ACAS terminology. Related to my comment above, it'd be great if we could either use the Protocol class here, or an Abstract class

acasclient/protocol.py
brianbolt: Yeah, this mostly makes sense to me.  A few notes:

- Protocols and experiments do actually have interactions with itx states and values but are unused for years
- We should decide where exactly we want the abstractions to go because right now ls_states and ls_labels ...etc. live on the LsThing class so things like "get_preferred_label" are currently on the LsThing class https://github.com/mcneilco/acasclient/blob/1b638bfae9a78a4d98c1537164ada3d0ad17e7f4/acasclient/lsthing.py#L1170

bffrost: We do already have `AbstractThing`, `AbstractLabel`, `AbstractState` and `AbstractValue`.

I think we could move the `get_preferred_label` and the ls_states, ls_labels handling within `as_dict` up into AbstractThing.


I have some clear ideas about how we could refactor SimpleLsThing also. Let's separate this refactor I'm proposing from this PR. I can open up a PR for it.

bffrost (approved):
Aside from the nit about an unused `ACAS_LSTHING` constant, this LGTM. Let's spin off the refactor to a separate PR, especially since this code is already tested through the migration.

acasclient/protocol.py
brianbolt: Ok yeah makes sense.  May end up being related but I started to pull additional fields up to the base model in this branch since they were shared fields.  Was also trying to simplify some of the initialization methods: https://github.com/mcneilco/acasclient/compare/DISCO-1604...DISCO-1604-protocol-refactor

PR #168, opened 2024-11-26
ACAS-797: Fix error when code_origin is explicitly None

bffrost:
## Description
Custom ACAS code is running into errors with latest acasclient where some CodeValues are being instantiated from ls_value objects with null / None code_origin. While unexpected, we should allow this at the acasclient level.

## How Has This Been Tested?
- Wrote automated test, confirmed error, then fixed it

## Extra Information
I'm planning to tag a new release of acasclient (`2024.4.0`) after merging this.


brianbolt (approved):
LGTM

PR #182, opened 2026-03-23
ACAS-956 extension: Add pytest and config-aware checker

bffrost:
## Description

A longstanding issue with our automated test coverage is that we can only test our default configs. This is a heavily Claude-fueled attempt at a solution. It adds:
- Utility to fetch `client` configs only from where they're already exposed by acas and parse them
- Added pytest as a testing dependency
- Added custom pytest decorator to mark tests as requiring certain configs to run, otherwise they're skipped

## How Has This Been Tested?

- Confirmed the test runs when local instance is running with `uniqueNotebook=true`
```
(.venv) ➜  acasclient git:(ACAS-956-3) pytest tests/test_acasclient.py::TestCmpdReg::test_019_duplicate_notebook_page
============================================== test session starts ===============================================
platform darwin -- Python 3.10.20, pytest-9.0.2, pluggy-1.6.0
rootdir: /Users/frost/Projects/ACAS/acasclient
collected 1 item                                                                                                 

tests/test_acasclient.py .                                                                                 [100%]

=============================================== 1 passed in 5.12s ================================================
```
- Confirmed the test skips when local instance has it false
```
(.venv) ➜  acasclient git:(ACAS-956-3) pytest tests/test_acasclient.py::TestCmpdReg::test_019_duplicate_notebook_page
============================================== test session starts ===============================================
platform darwin -- Python 3.10.20, pytest-9.0.2, pluggy-1.6.0
rootdir: /Users/frost/Projects/ACAS/acasclient
collected 1 item                                                                                                 

tests/test_acasclient.py s                                                                                 [100%]

=============================================== 1 skipped in 0.47s ===============================================
```

### client-only configs
The mechanism by which I'm fetching configs is limited to only client-side. We have many potential test cases for server-side configs, but want to avoid the security risk of exposing server-side configs via the frontend.
Some options I considered:
- Adding a special environment variable such as (`TEST_MODE`) and adding a new node API route that exposes the full configs only when that's true, then using that
- Having the pytest helper run `docker compose exec` to fetch the config file contents

### Managing "alternative config collections"
I have a POC working with docker compose override files to override config-related env vars. These could serve as small "libraries" of alternate configs to run tests with, but they would be checked into the `acas` repo and thus separated somewhat from the test suite here.

`docker-compose.test-unique-notebook.yml`
```
version: '2'

# Docker Compose override for testing with uniqueNotebook=true
#
# Usage:
#   docker-compose -f docker-compose.yml -f docker-compose.test-unique-notebook.yml up -d
#
# This enables the compound registration setting that enforces unique notebook pages
# across all lots in the database.

services:
  acas:
    environment:
      # Override the uniqueNotebook setting to true
      - ACAS_CLIENT_CMPDREG_SERVERSETTINGS_UNIQUENOTEBOOK=true
```

Then running the CI tests in a matrix like this: https://github.com/mcneilco/acasclient/compare/ACAS-956-2...ACAS-956-3?expand=1

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,0 +1,80 @@
+"""
+Pytest configuration for acasclient tests.
+
+Provides:
+- requires_config marker for config-dependent tests
+- Client config fetching from ACAS /conf/conf.js endpoint
+"""
+
+import json
+import pytest
+import requests
+
+
+# Cached config - fetched once per session
+_cached_config = None
+
+
+def get_nested_value(d, path):
+    """
+    Get a nested value from a dictionary using dot notation.
+
+    Args:
+        d: Dictionary to search
+        path: Dot-separated path (e.g., 'cmpdreg.serverSettings.uniqueNotebook')
+
+    Returns:
+        The value at the path, or None if not found
+    """
+    keys = path.split('.')
+    current = d
+    for key in keys:
+        if not isinstance(current, dict):
+            return None
+        current = current.get(key)
+        if current is None:
+            return None
+    return current
+
+
+def get_client_config(base_url="http://localhost:3000"):
+    """
+    Fetch client config from ACAS /conf/conf.js endpoint.
+
+    Cached after first call. Returns empty dict on error (tests will skip).
+    """
+    global _cached_config
+    if _cached_config is None:
+        try:
+            response = requests.get(f"{base_url}/conf/conf.js", timeout=5)
+            response.raise_for_status()
+            js_content = response.text
+            # Strip "window.conf=" prefix and parse JSON
+            json_str = js_content.replace("window.conf=", "", 1).rstrip(";")
+            _cached_config = json.loads(json_str)
+        except Exception as e:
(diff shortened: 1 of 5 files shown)


tests/conftest.py
```diff
@@ -0,0 +1,80 @@
...
+def get_client_config(base_url="http://localhost:3000"):
+    """
+    Fetch client config from ACAS /conf/conf.js endpoint.
+
+    Cached after first call. Returns empty dict on error (tests will skip).
+    """
+    global _cached_config
+    if _cached_config is None:
+        try:
+            response = requests.get(f"{base_url}/conf/conf.js", timeout=5)
```
brianbolt, line 49: This would only work for client side configs. Not server only configs. Not sure the best way to remedy that. Ideally we exposed all configs to client except for secrets but its not the way ACAS was designed.

brianbolt (commented):
I think its a good idea and is a nice addition but there is a caveat that is a blocker for some of the configs we have as noted in the review.

PR #183, opened 2026-03-25
ACAS-928: Cover additional special characters in search test

bffrost:
## Description

Original fix missed some locations which caused certain special characters to not work. Added a test that covers them to confirm the fix.

## How Has This Been Tested?

```
(.venv) ➜  acasclient git:(release/2026.2.x) ✗ python -m unittest -v tests.test_acasclient.TestAcasclient.test_057_get_protocol_expt_special_characters_in_names
test_057_get_protocol_expt_special_characters_in_names (tests.test_acasclient.TestAcasclient)
Test getting protocols and experiments with special characters in their names. ... ok
Successfully deleted all experiments
Successfully deleted all cmpdreg bulk load files
Successfully deleted all salts
Successfully deleted all projects (except Global)

----------------------------------------------------------------------
Ran 1 test in 5.548s

OK
```

### DIFF
--- a/tests/test_acasclient.py
+++ b/tests/test_acasclient.py
@@ -3009,14 +3009,17 @@ class TestAcasclient(BaseAcasClientTest):
     @requires_basic_cmpd_reg_load
     def test_057_get_protocol_expt_special_characters_in_names(self):
         """Test getting protocols and experiments with special characters in their names."""
-        protocol_name = "test/protocol.name !@#$%^&*(special/chars)"
-        experiment_name = "test/experiment.name !@#$%^&*(special/chars)"
+        protocol_name = "test/protocol.name !@#[%]$%^&*(special/chars)"
+        experiment_name = "test/experiment.name !@#[%]$%^&*(special/chars)"
         file_to_upload = get_basic_experiment_load_file(self.tempdir, protocol_name=protocol_name, experiment_name=experiment_name)
         response = self.client.\
             experiment_loader(file_to_upload, "bob", False)
         # Search for the protocol by name
         res = self.client.protocol_search(protocol_name)
         self.assertEqual(len(res), 1)
+        # Search for the protocol by substring
+        res = self.client.protocol_search("[")
+        self.assertEqual(len(res), 1, "Expected to find protocol when searching with partial special characters")
         # Get the protocol by name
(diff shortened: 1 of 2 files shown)


tests/test_acasclient.py
```diff
@@ -3009,14 +3009,17 @@ def test_056_additional_assay_scientists(self):
     @requires_basic_cmpd_reg_load
     def test_057_get_protocol_expt_special_characters_in_names(self):
         """Test getting protocols and experiments with special characters in their names."""
-        protocol_name = "test/protocol.name !@#$%^&*(special/chars)"
-        experiment_name = "test/experiment.name !@#$%^&*(special/chars)"
+        protocol_name = "test/protocol.name !@#[%]$%^&*(special/chars)"
+        experiment_name = "test/experiment.name !@#[%]$%^&*(special/chars)"
```
brianbolt: Wondering if it would make sense to have the special character as a variable, then prepend the protocol and experiment names.  Then below you can split the string and test each one?

brianbolt (changes requested):
LGTM

tests/test_acasclient.py
```diff
@@ -3009,14 +3009,21 @@ def test_056_additional_assay_scientists(self):
...
         file_to_upload = get_basic_experiment_load_file(self.tempdir, protocol_name=protocol_name, experiment_name=experiment_name)
         response = self.client.\
             experiment_loader(file_to_upload, "bob", False)
+        expt_code = response['results']['experimentCode']
+        protocol_code = self.client.get_experiment_by_code(expt_code)['protocol']['codeName']
         # Search for the protocol by name
         res = self.client.protocol_search(protocol_name)
         self.assertEqual(len(res), 1)
+        # Search for the protocol by substring
+        for char in protocol_name:
```
brianbolt: This is going to search special and non special characters and repeat multiple of the same searches . e.g. "s" shows up 3 times.  Can you seperate the special characters as a variable, then template them into the protocol/experiments names and then loop over just the special characters?

brianbolt (changes requested):

tests/test_acasclient.py
bffrost: Yep, did this.

brianbolt (approved):
Thanks. LGTM

